PR #16, opened 2014-05-23
[winrt support] Adding BOOST_NO_ANSI_APIS for when targeting the Windows Runtime.

stgates:
Uses the predef added for detection of the Windows Runtime. Also added a year to the copyright for my previous change.


eldiener:
The undocumented but used macro BOOST_NO_ANSI_APIS means that Windows does not have Ansi versions of the Windows API but only their corresponding wide character versions, ie does not have versions of their APIs ending in 'A' but only the equivalent version ending in 'W', such as having 'CreateFileW' but not 'CreateFileA'. It also looks like predef has no dependencies on other Boost libraries so using predef in Boost.config does not add on any additional dependencies.

jzmaddock:
Well yes it does, it adds predef ;)

Seriously, I'd prefer not to include predef if we can get away with it, and this should be detectable from compiler defines anyway?

mauve:
Here is a similar PR but without adding a dependency on Boost.Predef: https://github.com/boostorg/config/pull/80

jzmaddock:
Merged #80 instead, closing this one.

PR #40, opened 2014-09-26
Make std lib detection a bit more lightweight

igaztanaga:
...on environments where cstddef is enough to detect BOOST_STDLIB_CONFIG. Backup to `<utility>` (which has become heavier in C++11, as it needs traits for move_if_noexcept..) if std lib has not been detected.

The patch splits the detection in an auxiliary header and includes it twice in case cstddef is not enough.


jzmaddock:
Update, issue: https://github.com/boostorg/config/pull/58 reintroduces a dependency on <ultility> and we're trying to figure out what to do about it :(

PR #47, opened 2015-02-17
Update version.hpp

rpavlik:
Fix typo in comment, and also clean up capitalization and overall meaning.

(My header file doesn't change every time there's a new boost release, but it would if I upgraded.)


jzmaddock:
I've merged to develop, PR should close when I merge to master, John.

PR #57, opened 2015-04-22
add test maco for feature std::integer_sequence

olk:
- std::integer_sequence is a C++14 feature
- SD-6 macro __cpp_lib_integer_sequence is used to test this feature
- __cpp_lib_integer_sequence is defined in header <utility>
- if std::integer_sequence is not available, BOOST_NO_CXX14_INTEGER_SEQUENCE is defined

### DIFF
--- a/include/boost/config/compiler/clang.hpp
+++ b/include/boost/config/compiler/clang.hpp
@@ -8,6 +8,9 @@
 
 // Clang compiler setup.
 
+// __cpp_lib_integer_sequence is defined in header <utility>
+#include <utility>
+
 #define BOOST_HAS_PRAGMA_ONCE
 
 // Detecting `-fms-extension` compiler flag assuming that _MSC_VER defined when that flag is used.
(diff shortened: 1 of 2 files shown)


olk:
sorry - this patch modifies only GCC and Clang

jzmaddock:
Oliver, I'm sorry but I'm going to reject this as is, because:

* This is a std lib feature not a compiler one, macros should be moved appropriately.
* The macro should be defined in all config headers under /stdlib/ where this is unsupported (or unknown whether it is).  Each should have a section at the end already for "things that are always defined".
* We need a test case and docs.

Please see also http://www.boost.org/doc/libs/1_57_0/libs/config/doc/html/boost_config/guidelines_for_boost_authors.html#boost_config.guidelines_for_boost_authors.adding_new_defect_macros

If you don't submit a revised PR by the weekend, I might take care of this anyway, especially if you remind me!

PR #60, opened 2015-05-03
Update version check for VisualStudio 2015 RC.

Beman:
Looking at STL's list, I didn't see anything else requiring Boost.Config changes.

See http://blogs.msdn.com/b/vcblog/archive/2015/04/29/c-11-14-17-features-in-vs-2015-rc.aspx

--Beman


jzmaddock:
Thanks!

PR #61, opened 2015-05-05
Add more constexpr tests

MarcelRaad:
MSVC 14 RC currently passes the constexpr tests, but fails to compile Boost if BOOST_NO_CXX11_CONSTEXPR is not defined. Added three tests that clang can compile in C++11 mode, but MSVC cannot.


jzmaddock:
Thanks!

PR #64, opened 2015-06-02
adding toolset for xlcpp

camorton2:
adding a toolset 'xlcpp' for the new IBM compilers based on clang, old toolset 'vacpp' will remain for older IBM compilers not based on clang

### DIFF
--- a/include/boost/config/compiler/xlcpp.hpp
+++ b/include/boost/config/compiler/xlcpp.hpp
@@ -1,0 +1,271 @@
+// (C) Copyright Douglas Gregor 2010
+//
+//  Use, modification and distribution are subject to the
+//  Boost Software License, Version 1.0. (See accompanying file
+//  LICENSE_1_0.txt or copy at http://www.boost.org/LICENSE_1_0.txt)
+
+//  See http://www.boost.org for most recent version.
+
+//  compiler setup for IBM XL C/C++ for Linux (Little Endian) based on clang.
+
+#define BOOST_HAS_PRAGMA_ONCE
+
+// Detecting `-fms-extension` compiler flag assuming that _MSC_VER defined when that flag is used.
+#if defined (_MSC_VER) && (__clang_major__ > 3 || (__clang_major__ == 3 && __clang_minor__ >= 4))
+#   define BOOST_HAS_PRAGMA_DETECT_MISMATCH
+#endif
+
+// When compiling with clang before __has_extension was defined,
+// even if one writes 'defined(__has_extension) && __has_extension(xxx)',
+// clang reports a compiler error. So the only workaround found is:
+
+#ifndef __has_extension
+#define __has_extension __has_feature
+#endif
+
+#if !__has_feature(cxx_exceptions) && !defined(BOOST_NO_EXCEPTIONS)
+#  define BOOST_NO_EXCEPTIONS
+#endif
+
+#if !__has_feature(cxx_rtti) && !defined(BOOST_NO_RTTI)
+#  define BOOST_NO_RTTI
+#endif
+
+#if !__has_feature(cxx_rtti) && !defined(BOOST_NO_TYPEID)
+#  define BOOST_NO_TYPEID
+#endif
+
+#if defined(__int64) && !defined(__GNUC__)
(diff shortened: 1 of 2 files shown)


jzmaddock:
What's the difference between this and the regular clang.hpp?  Is there any reason why that one can't be used as is?

Either way the selection logic in select_compiler.hpp will need changing.

camorton2:
There's no difference other than the name of the compiler.
So I can select clang.hpp with the xlcpp toolset in select_compiler.hpp ?

camorton2:
I created a new file because its likely we might need to change something as we continue testing - we don't have all the clang features supported.

jzmaddock:
On 02/06/2015 19:43, Catherine Morton wrote:
>
> I created a new file because its likely we might need to change 
> something as we continue testing - we don't have all the clang 
> features supported.
>

Understood, but as most of that file just uses has_feature etc, should 
anything unsupported return false and do the right thing? Seems like 
it's worth a try at least?  But whatever, I'll go with whatever works 
best.  The main danger with a separate file is that it gets out of date 
/ under-maintained I guess.

camorton2:
I made the changes to include/boost/config/select_compiler_config.hpp

jzmaddock:
Merged to *develop*.  Closing manually.

jzmaddock:
This isn't quite working:

Looking at: http://www.boost.org/development/tests/develop/developer/output/IBM%20XL%20C++%20Compiler%20For%20Linux%2C%20V13-1-1-boost-bin-v2-libs-config-test-config_info-test-vacpp-release.html  I see that clang.hpp is still being selected.

The compiler also appears to be defining __SIZEOF_INT128__ without actually supporting __int128 which causes several sets of tests to fail, for example: http://www.boost.org/development/tests/develop/developer/output/IBM%20XL%20C++%20Compiler%20For%20Linux%2C%20V13-1-1-boost-bin-v2-libs-config-test-config_test-test-vacpp-release.html.  I guess we can just unconditionally disable the code that activates __int128 support once the correct config file is selected.  Do you have a fix that will cause the correct one to be selected?

camorton2:
Those tests were still using the vacpp toolset, we haven't enabled the xlcpp toolset yet

jzmaddock:
I'm not sure I follow - if you're testing Git-develop then the changes in this PR should be in effect?

camorton2:
Its a chicken/egg thing.   I created 3 pull requests to get the new toolset configured, but I can't enable it on our regular run until they are all in place.... Unless I misunderstand, the run script for the tests downloads a fresh copy of the source so it won't have all of my pull requests.   Our plan was to keep running with the previous toolset until all the pull requests are closed, then switch over and continue working out problems as we investigate failures.

jzmaddock:
> Its a chicken/egg thing. I created 3 pull requests to get the new 
> toolset configured, but I can't enable it on our regular run until 
> they are all in place.... Unless I misunderstand, the run script for 
> the tests downloads a fresh copy of the source so it won't have all of 
> my pull requests. Our plan was to keep running with the previous 
> toolset until all the pull requests are closed, then switch over and 
> continue working out problems as we investigate failures.
>
I suspect we may be talking at cross-purposes, but it shouldn't matter 
at all which Boost.Build toolset is used (if that's the other PR), 
especially as most users won't be using Boost.Build when they *use* 
Boost.  Wait, compiler pre-defines *both* __clang__ and __ibmxl__ yes?  
If that's the case then __clang__ is seen/configured first in 
select_compiler_config.hpp and so your patch has no effect?   If that's 
the case let me know and I'll make the "obvious" fix.  Thanks, John.

PS defining __SIZEOF_INT128__ without actually supporting __int128 is 
still a compiler bug IMO ;)

camorton2:
Thanks for pointing that out, please fix it for me.

I agree about the compiler bug having __SIZEOF_INT128__  defined without actually supporting __int128 - I've reported the defect.

jzmaddock:
Tentatively fixed in https://github.com/boostorg/config/commit/046b9d95e0d359f8f63eb6216e1e8345ecca0569, need to let the tests cycle to see.

PR #68, opened 2015-07-20
Update for the VS14 RTM standard library

MarcelRaad:
As mentioned by Stephan T. Lavavej here:
http://lists.boost.org/Archives/boost/2014/11/217698.php
BOOST_NO_CXX11_ADDRESSOF shouldn't be necessary anymore for MSVC 14 RTM. The unit test passes.


jzmaddock:
Thanks for keeping on top of this!

PR #70, opened 2015-08-11
Update for MSVC14's _HAS_AUTO_PTR_ETC

MarcelRaad:
If _HAS_AUTO_PTR_ETC is defined to 0, MSVC14's standard library implements C++17's N4190, thus removing std::auto_ptr.


eldiener:
The first correction from:

#if _CPPLIB_VER == 650

to

#if defined(_CPPLIB_VER) && (_CPPLIB_VER == 650)

is unnecessary. An undefined macro is always 0 in an #if statement.

MarcelRaad:
Yes it's unnecessary, but at least MSVC issues a warning (C4668) when using undefined preprocessor symbols and it's checked in all other places in this file. But I actually didn't mean to commit this as it's unrelated :-\

jzmaddock:
Thanks for this.

PR #71, opened 2015-08-27
Use __has_attribute instead of gcc version in clang

alkino:


jzmaddock:
The 

#ifndef __has_cpp_attribute

branch block should be:

#ifndef __has_attribute

?

alkino:
Hum sorry for that. Fixed now.

morinmorin:
The proposed detection code suffers from macro interference on not-so-recent versions of Clang (up to ver. 3.2): with `#define unused foo`, `__has_attribute(unused)` gets expanded to `__has_attribute(foo)`. The old detection code looks a bit awkward, but it's free from macro interference. 

Or rather, as Clang has supported the `unused` attribute since its first release, we can completely remove the detection code. I'll make a PR for this later.

I'll also make a PR that adds double underscores in
```
#  define BOOST_ATTRIBUTE_UNUSED __attribute__((unused))
```

P.S.
Here is history of `__has_attribute` on Clang.
- Clang 2.9 introduced `__has_attribute`. (This version also introduced `__GNUC__` as 4.)
- Clang 3.1 started to accept names with double underscores (`__xxxx__`) in `__has_attribute`.
- Clang 3.3 started to avoid macro expansion in `__has_attribute`.

alkino:
Thanks for pointing it

morinmorin:
Thanks for your contribution and heads-up too, Nicolas!

PR #74, opened 2015-09-03
Remove the detection code for 'unused' attribute on Clang

morinmorin:
Clang has supported the 'unused' attribute since the first release.

### DIFF
--- a/include/boost/config/compiler/clang.hpp
+++ b/include/boost/config/compiler/clang.hpp
@@ -260,9 +260,8 @@
 #  define BOOST_NO_CXX14_DIGIT_SEPARATORS
 #endif
 
-#if __has_attribute(unused)
-#  define BOOST_ATTRIBUTE_UNUSED __attribute__((unused))
-#endif
+// Clang has supported the 'unused' attribute since the first release.
+#define BOOST_ATTRIBUTE_UNUSED __attribute__((unused))
 
 #ifndef BOOST_COMPILER
 #  define BOOST_COMPILER "Clang version " __clang_version__


jzmaddock:
Thanks, that's much simpler!

morinmorin:
This PR turns on `__attribute__((unused))` on all versions of Clang. But Clang 1.0 and 1.1 support the 'unused' attribute only for **variables and functions**; they do not accept the 'unused' attribute for **types** and emit warnings if we put the attribute. 

IMHO, Clang 1.0 and 1.1 are old enough to drop the support, but if you like, I'll of course revert the detection code to the original one: 
```
#if defined(__GNUC__) && (__GNUC__ >= 4)
```

Sorry about the breakage.

jzmaddock:
On 10/09/2015 16:26, morinmorin wrote:
>
> This PR turns on |__attribute__((unused))| on all versions of Clang. 
> But Clang 1.0 and 1.1 support the 'unused' attribute only for 
> *variables and functions*; they do not accept the 'unused' attribute 
> for *types* and emit warnings if we put the attribute.
>
> IMHO, Clang 1.0 and 1.1 are old enough to drop the support,
>

+1, I seriously doubt anyone is using clang v1, and I suspect if they 
are, they would have other issues anyway?

Suggest we leave it as it is for now, John.

morinmorin:
> +1, I seriously doubt anyone is using clang v1, and I suspect if they are, they would have other issues anyway?

Totally agreed.

> Suggest we leave it as it is for now, John.

OK, thanks!

Regards,
Michel

PR #75, opened 2015-09-03
Guard attribute names with double underscores

morinmorin:
Both gcc and clang (since their early versions like gcc-2.9.5 and clang-1.0) accept attribute names with double underscores to avoid macro interference. So it's better to guard attribute names with double underscores, but there are unguarded ones in `clang.hpp` and `gcc.hpp`: 
```
#  define BOOST_ATTRIBUTE_UNUSED __attribute__((unused))
```
This PR fix this.


jzmaddock:
Thanks for that!

PR #77, opened 2015-09-18
Make cstdint.hpp work on QNX 6.6

mauve:
Before this commit we were using some QNX special extensions which have been deprecated, let's not use them unless QNX is built with extensions enabled.


jzmaddock:
Merged to develop.

PR #80, opened 2015-10-04
Do not use Windows ANSI APIs when targetting WinRT

mauve:
(This is a similar PR as https://github.com/boostorg/config/pull/16 but with the difference that it doesn't pull in a dependency on Boost.Predef)

When targeting WinRT (phone, store, universal) we do not want to use ANSI versions of Windows APIS as a lot of the old win32 ansi apis are missing, therefore let's skip them altogether.


jzmaddock:
That looks better, thanks.

PR #83, opened 2015-12-18
Fix Trac 11852: CUDA in C++03 Mode and __float128

ax3l:
Fix trac issue
  https://svn.boost.org/trac/boost/ticket/11852

Similar to
     https://svn.boost.org/trac/boost/ticket/8048

`__float128` is still unsupported when compiling with nvcc *in non-C++11 mode* (tested with `CUDA 7.5.18`). First noticed with the latest release (`1.60.0`) and `GCC 4.8.5` but should affect all previous releases depending on used modules.

~~In my case, I triggered it with the components
  `program_options regex filesystem system thread math_tr1`
enabled.~~

The reason this is triggered in headers is that `boost 1.60.0` introduced `__float128` type traits in [type_traits/is_floating_point.hpp](https://github.com/boostorg/type_traits/blob/boost-1.60.0/include/boost/type_traits/is_floating_point.hpp#L24-L26) which were not implemented in the previous release [1.59.0](https://github.com/boostorg/type_traits/blob/boost-1.59.0/include/boost/type_traits/is_floating_point.hpp).

### DIFF
--- a/include/boost/config/compiler/clang.hpp
+++ b/include/boost/config/compiler/clang.hpp
@@ -57,16 +57,25 @@
 #define BOOST_HAS_LONG_LONG
 
 //
-// We disable this if the compiler is really nvcc as it
-// doesn't actually support __int128 as of CUDA_VERSION=5000
+// We disable this if the compiler is really nvcc with C++03 as it
+// doesn't actually support __int128 as of CUDA_VERSION=7500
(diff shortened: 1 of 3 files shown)


jzmaddock:
Thanks for looking into this!

ax3l:
no problem, thank you!

PR #90, opened 2016-05-25
Added generalized build configuration test for setting any requiremen…

eldiener:
…t when true or false and changed the 'requires' test to work in terms of the generalized test. I have not documented the generalized test but am willing to do so. Also I do not know if there is any Boost.config test for 'requires' so I have not attempted to test the changed functionality in a formal Boost.config test, although I have tested it out from another library on which I am working to ascertain that 'requires' and the new 'requires.set.properties' rules work accordingly.

### DIFF
--- a/checks/config.jam
+++ b/checks/config.jam
@@ -7,6 +7,11 @@ import modules ;
 
 rule requires ( names + )
 {
+   return [ requires.set.properties $(names) : : <build>no ] ;
+}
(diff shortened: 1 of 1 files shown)


eldiener:
I realize that this does not work properly with multiple targets setting iftrue and iffalse requirements, and therefore I am going to close this PR.

PR #91, opened 2016-05-29
Add autolinking support for clang-cl

ismail:
clang-cl emulates all of the cl.exe options so we can fake it as BOOST_MSVC as far as autolinking is concerned and undef it at the end. Tested with boost 1.61.0 with clang 3.9 trunk.


ismail:
Doesn't seem to work with boost::thread, closing for now :/

PR #99, opened 2016-11-14
Update for Visual Studio 15 Preview 5

MarcelRaad:
- Aggregate NSDMI and relaxed constexpr are supported
- _MSC_VER is 1910

All Config tests are passing and all projects using Boost I have access to compile and run fine.

### DIFF
--- a/include/boost/config/compiler/visualc.hpp
+++ b/include/boost/config/compiler/visualc.hpp
@@ -190,6 +190,13 @@
 #  define BOOST_NO_CXX11_CONSTEXPR
 #endif
 
+// C++14 features supported by VC++ 15 Preview 5
+//
+#if (_MSC_VER < 1910)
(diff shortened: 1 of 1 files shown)


Beman:
I've had a number of conversions with Microsoft developers about these config macros. Two of them where the actual devs who have been processing bug reports for these features. If I recall correctly, there are a number of bug fixes that did not make it into the preview 5 release, but are supposed to be in the RC. Since the RC was announced this morning, you might want to download a copy and base the patch on the RC _MSC_VER value. I'm tied up and won't be able to do that myself for awhile.

Thanks for working on this
--Beman

MarcelRaad:
Thanks @Beman ! I'll do that later today.

MarcelRaad:
Updated the version number. _MSC_VER is still 1910. Everything still works.

PR #105, opened 2016-12-11
defect macro for C++17 feature `std::apply()`

olk:


jzmaddock:
That all looks fine (thanks!), but as I'm short of time... could you possibly resolve the conflicts with develop?

olk:
I've no idea why git think it has merge conflicts, commit seams correct for both files.
Unfortunately I've no write permissions for boost.config

PR #110, opened 2017-01-05
Use corrent name for libraries built with MS VS 2017 RC.

o01eg:
Boost built with MS VS 2017 RC has -vc150 suffix.

### DIFF
--- a/include/boost/config/auto_link.hpp
+++ b/include/boost/config/auto_link.hpp
@@ -161,10 +161,15 @@ BOOST_LIB_VERSION:    The Boost version, in the form x_y, for Boost version x.y.
      // vc12:
 #    define BOOST_LIB_TOOLSET "vc120"
 
+#  elif defined(BOOST_MSVC) && (BOOST_MSVC < 1910)
+
+     // vc14:
+#    define BOOST_LIB_TOOLSET "vc140"
+
 # elif defined(BOOST_MSVC)
 
-   // vc14:
-#  define BOOST_LIB_TOOLSET "vc140"
+   // vc15:
+#  define BOOST_LIB_TOOLSET "vc150"
 
 #  elif defined(__BORLANDC__)
 


eldiener:
Is 1910 the minimum _MSC_VER for VS 2017 ?

jhunold:
Yes, the official version is 1910. I'm not sure if the correct library suffix should be -141 instead, as the compiler itself is still ABI-compatible with vc14. And the Boost.Build support for this is still flaky. I'm not sure I'll have the time to get it stable, though.

eldiener:
Then maybe we should just wait until VS 2017 is officially released instead of hacking away at it and not really knowing what it will be until it is released.

jzmaddock:
Do we know if Microsoft are calling it VC15 or VC14.1?

o01eg:
It defined in msvc.jam from Boost.Build in https://github.com/boostorg/build/commit/ae5e63a1315493dfb2fc60c4d537674fa637f898 which was released in 1.63.0 version.

KindDragon:
> Boost built with MS VS 2017 RC has -vc150 suffix.

I think libraries should be named with suffix `-vc140` because VS2017 fully backward compatible with VS2015.

KindDragon:
> Do we know if Microsoft are calling it VC15 or VC14.1?

Visual Studio version will be 15 (https://visualstudio.uservoice.com/forums/121579-visual-studio-ide/suggestions/13536840-adopt-a-simple-and-obvious-versioning-scheme-for-t), but CRT version will be 14.1 (https://build2.org/article/notes-visual-studio-15.xhtml)

o01eg:
Boost build already generates vc150 suffix.

OlafvdSpek:
> I think libraries should be named with suffix -vc140 because VS2017 fully backward compatible with VS2015.

Does vc140 support std::string_view too?
vc141 does..

KindDragon:
> Boost build already generates vc150 suffix.

Boost Build can't compile Boost with VS2017 in a current state.

o01eg:
@KindDragon It has initial support and with some patches I built it: https://github.com/o01eg/freeorion-sdk

KindDragon:
@o01eg Yes it possible, but in general user wouldn't fo it. So we can change suffix if need it

jzmaddock:
OK let's try this for now, and change later if needed.

KindDragon:
So user can define `BOOST_LIB_TOOLSET ="vc140"` to use `vc140` libraries with VS2017?

jzmaddock:
> So user can define |BOOST_LIB_TOOLSET ="vc140"| to use |vc140| 
> libraries with VS2017?
>

Yes.

PR #113, opened 2017-02-02
Support BOOST_GCC guards in BOOST_WORKAROUND

glenfe:
BOOST_GCC is absent among the convenience guards present for BOOST_WORKAROUND.


glenfe:
There are guards in workaround.hpp for BOOST_INTEL and BOOST_MSVC (among other macros). This adds guards for BOOST_GCC as well so that we do not see warnings when compiling with -Wundef.

PR #114, opened 2017-02-04
Add BOOST_NO_CXX11_SFINAE_EXPR defect macro

glenfe:
This proposes a new BOOST_NO_CXX11_SFINAE_EXPR defect macro whose absence will indicate support for C++11 SFINAE expressions. The test case shows C++11 code which will not compile with MSVC (any version) and g++4.7 (with -std=c++11) even when the existing BOOST_NO_SFINAE_EXPR macro is not defined.


glenfe:
John, let me know if you need more context (hopefully the test case is illustrative enough) or if I need to change any other files (I just updated the pull request with the Quick book documentation entry).

PR #116, opened 2017-02-07
Update boost_no_cxx11_sfinae_expr test to fail faster

glenfe:
Will now compile-fail for Intel C++ 13 instead of run-fail.


jzmaddock:
This is failing for almost all clang test runners, can you investigate?

Generally I tend to assume that clang is correct... but you never know.

glenfe:
That's clang in C++03 mode (i.e. with -std=c++11 or higher you won't see those errors). The test code uses C++11. I'll submit a pull request shortly.

PR #119, opened 2017-02-20
Fix toolset name for VS2017

KindDragon:
I think we should change VS2017 toolset name to vc141 before next Boost release.

Related to https://github.com/boostorg/build/pull/167

/cc @AndrewPardoe

### DIFF
--- a/include/boost/config/auto_link.hpp
+++ b/include/boost/config/auto_link.hpp
@@ -166,10 +166,10 @@ BOOST_LIB_VERSION:    The Boost version, in the form x_y, for Boost version x.y.
      // vc14:
 #    define BOOST_LIB_TOOLSET "vc140"
 
-# elif defined(BOOST_MSVC)
+#  elif defined(BOOST_MSVC)
 
-   // vc15:
-#  define BOOST_LIB_TOOLSET "vc150"
+     // Visual Studio 2017:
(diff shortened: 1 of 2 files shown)


jzmaddock:
Some comments/questions:

1) Boost.Build needs to change this first Config just matches whatever lib names are produced by BB.
2) Boost.Build needs to support the next visual studio release period before this means much of anything anyway.
3) Do we have official confirmation of what MS are calling this release?  What's the _MSC_VER version?

Thanks.

KindDragon:
> Boost.Build needs to change this first Config just matches whatever lib names are produced by BB.

PR  boostorg/build#167 should fix this

> Boost.Build needs to support the next visual studio release period before this means much of anything anyway.

Now BB supports VS2017 if VS is installed to default path or you run Boost Build from "Developer Command Prompt for VS 2017".

> Do we have official confirmation of what MS are calling this release? What's the _MSC_VER version?

It should be 1910. I think @AndrewPardoe can help us verify this.

AndrewPardoe:
VS2017 contains the VC 14.1 toolset, or vc141. Part of the confusion is that the compiler's internal version, 19.10, is five higher than the public version, 14.1. So _MSC_VER=1910, but the toolset is vc141.

jzmaddock:
This is now addressed in develop.

PR #124, opened 2017-02-28
add _WORKAROUND_GUARD macros for BOOST_XLCPP_ZOS

pgroke-dt:
I'd missed workaround.hpp. And since `__IBMCPP__` doesn't uniquely identify the z/OS compiler, I guess there should be a `_WORKAROUND_GUARD` for `BOOST_XLCPP_ZOS`...

Is there anything else that needs to be done for `BOOST_WORKAROUND()` support?

### DIFF
--- a/include/boost/detail/workaround.hpp
+++ b/include/boost/detail/workaround.hpp
@@ -90,6 +90,11 @@
 #else
 #define BOOST_GCC_WORKAROUND_GUARD 0
 #endif
+#ifndef BOOST_XLCPP_ZOS
+#define BOOST_XLCPP_ZOS_WORKAROUND_GUARD 1
+#else
+#define BOOST_XLCPP_ZOS_WORKAROUND_GUARD 0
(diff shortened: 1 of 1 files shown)


jzmaddock:
This won't do any harm - the _WORKAROUND_GUARD thing just suppresses warnings that may be thrown out by the preprocessor otherwise.  Unless BOOST_WORKAROUND is actually being used with BOOST_XLCPP_ZOS though it isn't really required either.

pgroke-dt:
Thanks for merging this! :)
I will need `BOOST_WORKAROUND` for a patch in Boost.DateTime where `BOOST_DATE_TIME_NO_LOCALE` is defined. (Well, "need" is a relative term - the conditions there use `BOOST_WORKAROUND` and I thought it'd be nicer if I added my condition with `BOOST_WORKAROUND` as well -- it is a workaround after all and I like "uniformity".)

PR #126, opened 2017-03-16
VS2017 version used in other places is vc1410

teeks99:


glenfe:
It almost feels like all the VS2017 discussions around this are going around in circles. This is what was proposed in https://github.com/boostorg/config/pull/119 too.

PR #127, opened 2017-03-19
Change msvc version to "141"

jzmaddock:

PR #128, opened 2017-03-19
Using 141 instead of 1410

teeks99:
This PR should *not* be accepted unless the [community consensus on the mailing list](http://lists.boost.org/Archives/boost/2017/03/233543.php) converges around option 2.


mloskot:
Related PR #127

PR #129, opened 2017-03-24
Add cmake support

pfultz2:
This adds cmake support for building, testing, and installing Boost.Config. This requires first installing [BCM](https://github.com/boost-cmake/bcm). To build, test, and install Boost.Config, just run:

```
mkdir build
cd build
cmake ..
cmake --build . --target
cmake --build . --target check
cmake --build . --target install
```

This also adds travis and appveyor which will test the library on every push.


jzmaddock:
Closing as obsolete - Config now has it's own CI, and there is a Boost-wide CMake effort.

PR #131, opened 2017-04-09
Introduce feature-test macro BOOST_NO_CXX17_STD_RANDOM_SHUFFLE

DanielaE:
The algorithm std::random_shuffle was marked as deprecated in C++14 and is removed from the upcoming C++17 standard. The name itself is still reserved and library implementations may continue to ship it. In this sense, support for std::random_shuffle must be regarded as purely optional and a feature-test macro is required to test for its presence.

Signed-off-by: Daniela Engert <dani@ngrt.de>

### DIFF
--- a/doc/html/boost_config/boost_macro_reference.html
+++ b/doc/html/boost_config/boost_macro_reference.html
@@ -3770,7 +3770,7 @@
       that describe C++17 features not supported</a>
 </h3></div></div></div>
 <p>
-        The following macros describe features in the 2016 ISO C++ standard, formerly
+        The following macros describe features in the 2017 ISO C++ standard, formerly
         known as C++1z, that are not yet supported by a particular compiler or library.
       </p>
 <div class="informaltable"><table class="table">
@@ -3815,7 +3815,18 @@
                 </p>
               </td>
 </tr>
-</tbody>
+<tr>
+<td>
+                <p>
+                  <code class="computeroutput"><span class="identifier">BOOST_NO_CXX98_RANDOM_SHUFFLE</span></code>
+                </p>
+              </td>
+<td>
+                <p>
+                  The compiler does no longer support <code class="computeroutput"><span class="identifier">std</span><span class="special">::</span><span class="identifier">random_shuffle</span><span class="special">()</span></code>. It was deprecated in C++14 and is removed from C++17.
+                </p>
+              </td>
+</tr></tbody>
 </table></div>
 </div>
 <div class="section">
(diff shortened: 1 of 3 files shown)


glenfe:
I wonder if this should instead be named 'BOOST_NO_CXX17_DEPRECATED_STD_RANDOM_SHUFFLE' or something similar. Or some convention established to indicate features deprecated rather than features not yet available.

DanielaE:
Good question, Glen! I'm in favour of differentiating between a feature not yet (or not sufficiently) available, and a feature being removed after deprecation. OTOH, there is BOOST_NO_AUTO_PTR which is repurposed now for the absence of the deprecated std::auto_ptr in msvc's library in c++17 mode, even though it is documented as a c++03 defect macro. But that shouldn't prevent us from looking for a better naming scheme for features removed from the standard.

jzmaddock:
On 15/04/2017 19:42, Glen Fernandes wrote:
>
> I wonder if this should instead be named 
> 'BOOST_NO_CXX17_DEPRECATED_STD_RANDOM_SHUFFLE' or something similar. 
> Or some convention established to indicate features deprecated rather 
> than features not yet available.
>

It's not just deprecated, it's removed.  And yes we have a few of these 
- std::unary/binary_function in particular are used all over Boost I 
suspect.

I like the BOOST_NO/BOOST_HAS convention because it keeps everything 
simple and reasonably obvious, but I guess if we were being systematic, 
this would be something like BOOST_<STD>_REMOVED_<FEATURE> ?

---
http://www.avg.com

DanielaE:
So, that would lead to either

- BOOST_NO_AUTO_PTR
- BOOST_NO_RANDOM_SHUFFLE
- BOOST_NO_BINDERS_ADAPTERS

or

- BOOST_CXX17_REMOVED_AUTO_PTR
- BOOST_CXX17_REMOVED_RANDOM_SHUFFLE
- BOOST_CXX17_REMOVED_BINDERS_ADAPTERS

What's your pick?

jzmaddock:
I've asked on the mailing list, we need to establish a convention once 
and for all for this.


---
http://www.avg.com

jzmaddock:
As per discussion on the list, can we change this one to BOOST_NO_CXX98_RANDOM_SHUFFLE, rationale:

* Name should reflect the std that introduced the feature.
* There's no precedent for including "STD_" in the macro name - everything is in std:: anyway.

Thanks!  John.

DanielaE:
Alright, John - here we go!

Ciao
   Dani

jzmaddock:
Merged, also added tests, and 2 more related macros: BOOST_NO_CXX98_BINDERS and BOOST_NO_CXX98_FUNCTION_BASE.  Moved docs for these into a new section for removed features.

DanielaE:
Thanks, John - these would have been the next pull-requests of mine 😀

PR #135, opened 2017-05-19
The library implementations in MSVC 14.0 and 14.1 have both version n…

DanielaE:
…umber V6.50:0009, and therefore _CPPLIB_VER is defined to 650. But the library in 14.0 does *not* have std::apply whereas the one in 14.1 *does*.

Signed-off-by: Daniela Engert <dani@ngrt.de>

### DIFF
--- a/include/boost/config/stdlib/dinkumware.hpp
+++ b/include/boost/config/stdlib/dinkumware.hpp
@@ -163,7 +163,7 @@
 #endif
 
 // C++17 features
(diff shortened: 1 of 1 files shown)


glenfe:
This is really unfortunate. We need a good way to disambiguate MSVC standard library versions going forward that is simple.

DanielaE:
Yeah, but I really hope this is just a one-time fluke.

jzmaddock:
Just so you all know, I'm trying to get the travis test matrix clean 
before I merge anything new...

---
http://www.avg.com

PR #136, opened 2017-05-19
define BOOST_NO_AUTO_PTR when building with libc++ and C++17

mclow:
New libc++ versions remove `std::auto_ptr` when building as C++17, unless the magic tag `_LIBCPP_ENABLE_CXX17_REMOVED_AUTO_PTR` is defined. Tell the rest of Boost that there's no `auto_ptr` in that case.


mclow:
Just to be clear - this is Peter Dimov's idea. I think it's great, but don't want to take credit.

PR #140, opened 2017-05-25
Merge develop to master

glenfe:


jzmaddock:
See #141 
I'm in the process of setting up CI testing for Boost.Config, and there's quite a churn from that at present.

PR #149, opened 2017-05-29
Add feature defect detection for C++17 structured bindings and inline variables

glenfe:


include/boost/config/compiler/visualc.hpp
```diff
@@ -218,6 +218,10 @@
 //
 #define BOOST_NO_TWO_PHASE_NAME_LOOKUP
 #define BOOST_NO_CXX11_SFINAE_EXPR
+// C++ 17:
+#define BOOST_NO_CXX17_STRUCTURED_BINDINGS
+#define BOOST_NO_CXX17_INLINE_VARIABLES
+#define BOOST_NO_CXX17_FOLD_EXPRESSIONS
```
jzmaddock: Is there a reason for unconditionally defining these here, rather than using the std feature macros?  Just curious, John.

glenfe: I had heard that VC has no intention to support the SD-6 macros, but this information might be out of date. I can update the PR if you think it worth it (up until 2017 they are not supported, at least).

jzmaddock: Lets ignore it for now then ;)

PR #150, opened 2017-05-29
Correct logic about BOOST_NO_CTYPE_FUNCTIONS

coypoop:
Now it will no longer apply for netbsd.

netbsd: https://nxr.netbsd.org/xref/src/include/ctype.h
freebsd: https://grok.dragonflybsd.org/xref/freebsd/include/ctype.h
dragonflybsd: https://grok.dragonflybsd.org/xref/dragonfly/include/ctype.h
openbsd: https://grok.dragonflybsd.org/xref/openbsd/include/ctype.h

### DIFF
--- a/include/boost/config/platform/bsd.hpp
+++ b/include/boost/config/platform/bsd.hpp
@@ -62,7 +62,7 @@
 //
 // The BSD <ctype.h> has macros only, no functions:
 //
-#if !defined(__OpenBSD__) || defined(__DragonFly__)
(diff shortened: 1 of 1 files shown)


jzmaddock:
Is it possible to version this change?  Clearly older BSD versions will still have the issue, but I don't know enough about how BSD is versioned...

coypoop:
I suppose it doesn't actually matter, since both options seem to work, I thought it applies to more than just netbsd at first.
sorry for the noise

PR #152, opened 2017-05-30
fix C++11 CUDA 8 nvcc compile time odeint error

psychocoderHPC:
fix https://svn.boost.org/trac/boost/ticket/13049

disable usage of noexcept with `BOOST_NO_CXX11_NOEXCEPT` for nvcc from CUDA 8

### DIFF
--- a/include/boost/config/compiler/nvcc.hpp
+++ b/include/boost/config/compiler/nvcc.hpp
@@ -30,3 +30,11 @@
 #if defined(_MSC_VER)
 #  define BOOST_NO_CXX11_CONSTEXPR
 #endif
+
+// A bug in nvcc (cudafe) of CUDA 8.0 prevents use of noexcept
+// https://svn.boost.org/trac/boost/ticket/13049
+// The issue will be fixed with CUDA 9.
+#if (__CUDACC_VER__ >= 80000) && (__CUDACC_VER__ < 80100)
+#   define BOOST_NO_CXX11_NOEXCEPT
+#endif
+


jzmaddock:
Is this only device code that is effected, or does it hit the host meta-compiler as well?

psychocoderHPC:
This bug effects host code ([see example](https://svn.boost.org/trac10/ticket/13049#Example)) which is compiled with nvcc out of a `*.cu` file.

jzmaddock:
Reproduced and fixed in https://github.com/boostorg/config/commit/9b8de65f1dae960ecfde689a99c83f39ba60e3f9

akors:
Hi! Unfortunately, your fix broke compilation with CUDA 9. The reason for that is that you use the symbol __CUDACC_VER__ for version checking, but the symbol has been deprecated in CUDA 9.

Line 64 in `crt/common_functions.h` of the CUDA toolkit now reads:
`#define __CUDACC_VER__ "__CUDACC_VER__ is no longer supported.  Use __CUDACC_VER_MAJOR__, __CUDACC_VER_MINOR__, and __CUDACC_VER_BUILD__ instead."
`
The fix seems simple enough, instead of checking for the whole version, you could check separately for __CUDACC_VER_MAJOR__ and __CUDACC_VER_MINOR__ .

psychocoderHPC:
@akors You are right, but the deprecation breaks all checks in `include/boost/config/compiler/nvcc.hpp`. 
Therefore the full file needs an update.

akors:
Ok, that's a shame. Here's the Trac ticket URL: https://svn.boost.org/trac10/ticket/13152

psychocoderHPC:
thx for opening the ticket: btw NVIDIA breaks all old boost versions with this change.

ax3l:
ouch, it might be worth opening a CUDA bug report as long as CUDA 9 is still in RC!

deprecating and removing this var are two things, this change in the header goes straight to "unannounced removal".

for the boost fix: is `__CUDACC_VER_MAJOR/MINOR/BUILD__` also available since CUDA 7.5? (I think yes.)

akors:
> ouch, it might be worth opening a CUDA bug report as long as CUDA 9 is still in RC!

Good idea! I just did that too, the internal link is https://developer.nvidia.com/nvidia_bug/1971621 , in case anyone has access to that.

psychocoderHPC:
I tested the beta version of CUDA 9.0067 with a little program:

```C++
#include <iostream>
#include <cuda_runtime.h>

int main()
{
        std::cout<<__CUDACC_VER__<<std::endl;
}
```
compiel and run
```
nvcc main.cu
./a.out
90067 #this is the output
```

This means that the define is available in the beta, we need to check the latest rc.

ax3l:
The problem appears with the second CUDA 9 release candidate, CUDA 9.0.103.

jzmaddock:
Can you guys please verify that https://github.com/boostorg/config/pull/175 fixes this?

Many thanks, John Maddock.

PR #159, opened 2017-06-30
Add address model to auto_link.hpp

pdimov:
This corresponds to boostorg/boost#147 and should be applied in tandem with it.

I am not sure whether the tagged layout needs the address model too, so I've left it untouched for now.


pdimov:
Superseded by #160.

PR #160, opened 2017-07-07
 Use architecture and address model in autolink

pdimov:
This corresponds to the superproject PR boostorg/boost#149 and should be applied in tandem with it.


pdimov:
I've merged boostorg/boost#149 so this should also be merged to update autolink to find the libraries.

pdimov:
Since I merged the arch-and-model superproject changes to master, I've taken the liberty of merging develop to master here. I initially cherry-picked just the commits in this PR, but the two remaining commits on develop address obvious issues, one of which fixes Travis, so I decided to include them as well.

PR #163, opened 2017-07-12
Fix references to headers in the docs

Lastique:
A few headers were moved to `boost/config/detail`. This PR updates the docs accordingly.


jzmaddock:
I've merged part of this, I'll regenerate the docs later as part of other pending changes.

PR #164, opened 2017-07-20
add BOOST_HASH_NO_EXTENSIONS to compile/cray.hpp

rsdale:
Avoid problems with multiple definitions of std::list using gnu headers.
	modified:   cray.hpp


danieljames:
This is probably the wrong solution, but you shouldn't set hash macros in config anyway.

rsdale:
I will look for a better solution.

I've got a problem with multiple definitions of std::list that seems 
related to the  _GLIBCXX_USE_CXX11_ABI macro.
I get that this is a hack, but it will allow my customers to compile 
their applications...and it's the best I can do today.

Thanks for your note.
--
regards,
rsd

PR #169, opened 2017-07-26
Update PGI C++ compiler support

dkolsen-pgi:
Remove an old PGI-specific workaround for intptr_t.  The workaround is no longer necessary and now causes compilation errors.

### DIFF
--- a/include/boost/cstdint.hpp
+++ b/include/boost/cstdint.hpp
@@ -367,9 +367,6 @@ namespace boost
 #include <stddef.h>
 #endif
 
-// PGI seems to not support intptr_t/uintptr_t properly. BOOST_HAS_STDINT_H is not defined for this compiler by Boost.Config.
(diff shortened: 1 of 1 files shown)


jzmaddock:
Is it possible to make the workaround PGI version checked, rather than removed altogether?

dkolsen-pgi:
It would be possible to have a version check, but it would not be useful.  The versions of the PGI compiler that needed this workaround are old enough that we (PGI) do not know of any customers still actively those versions.

jzmaddock:
On 27/07/2017 18:01, David Olsen wrote:
>
> It would be possible to have a version check, but it would not be 
> useful. The versions of the PGI compiler that needed this workaround 
> are old enough that we (PGI) do not know of any customers still 
> actively those versions.
>

Fair enough!

---
http://www.avg.com

PR #175, opened 2017-08-11
Update for CUDA version macro changes.

jzmaddock:


include/boost/config/compiler/nvcc.hpp
```diff
@@ -11,6 +11,15 @@
 #  define BOOST_COMPILER "NVIDIA CUDA C++ Compiler"
 #endif
 
+#if defined(__CUDACC_VER_MAJOR__) && defined(__CUDACC_VER_MINOR__) && defined(__CUDACC_VER_BUILD__)
+#  define BOOST_CUDA_VERSION __CUDACC_VER_MAJOR__ * 10000 + __CUDACC_VER_MINOR__ * 100 + __CUDACC_VER_BUILD__
```
ax3l: @jzmaddock careful, even CUDA 9 RC2 is already above 100 builds in `__CUDACC_VER_BUILD__`.

In my bug report I suggested to either increase for CUDA 9 build by a full order (*10) or to crop the build to two significant ints from the left during deprecation.

include/boost/config/compiler/nvcc.hpp
```diff
@@ -11,6 +11,15 @@
...
+#elif defined(__CUDACC_VER__)
```
ax3l: note: at least in CUDA 9 RC2 this var will be defined and `__CUDACC_VER__` *is* set to a string with an error message.

since the other 3 are defined above, this `elif` should not be taken but it's maybe still a bit risky.

jzmaddock:
On 14/08/2017 09:05, Axel Huebl wrote:
>
> *@ax3l* commented on this pull request.
>
> I would crop the |*_BUILD__| to two significant digits from the left
>
> ------------------------------------------------------------------------
>
> In include/boost/config/compiler/nvcc.hpp 
> <https://github.com/boostorg/config/pull/175#discussion_r132892843>:
>
> > @@ -11,6 +11,15 @@
>   #  define BOOST_COMPILER "NVIDIA CUDA C++ Compiler"
>   #endif
>   
> +#if defined(__CUDACC_VER_MAJOR__) && defined(__CUDACC_VER_MINOR__) && defined(__CUDACC_VER_BUILD__)
> +#  define BOOST_CUDA_VERSION __CUDACC_VER_MAJOR__ * 10000 + __CUDACC_VER_MINOR__ * 100 + __CUDACC_VER_BUILD__
>
> @jzmaddock <https://github.com/jzmaddock> careful, even CUDA 9 RC2 is 
> already above 100 builds in |__CUDACC_VER_BUILD__|.
>
> In my bug report I suggested to either increase for CUDA 9 build by a 
> full order (*10) or to crop the build to two significant ints from the 
> left during deprecation.
>

OK, or is there any reason to include the build number at all - perhaps 
we should just include the major and minor version numbers?

> ------------------------------------------------------------------------
>
> In include/boost/config/compiler/nvcc.hpp 
> <https://github.com/boostorg/config/pull/175#discussion_r132893153>:
>
> > @@ -11,6 +11,15 @@
>   #  define BOOST_COMPILER "NVIDIA CUDA C++ Compiler"
>   #endif
>   
> +#if defined(__CUDACC_VER_MAJOR__) && defined(__CUDACC_VER_MINOR__) && defined(__CUDACC_VER_BUILD__)
> +#  define BOOST_CUDA_VERSION __CUDACC_VER_MAJOR__ * 10000 + __CUDACC_VER_MINOR__ * 100 + __CUDACC_VER_BUILD__
> +#elif defined(__CUDACC_VER__)
>
> note: at least in CUDA 9 RC2 this var will be defined and 
> |__CUDACC_VER__| /is/ set to a string with an error message.
>
> since the other 3 are defined above, this |elif| should not be taken 
> but it's maybe still a bit risky.
>

I guess if __CUDACC_VER__ and __CUDACC_VER_MAJOR__ were introduced at 
the same time that branch may be redundant anyway.


---
http://www.avg.com

ax3l:
> OK, or is there any reason to include the build number at all - perhaps we should just include the major and minor version numbers?

Yes we should, CUDA 8.0 got after the initial public release (8.0.44) a bugfix release (8.0.61) and things can change between those. Even if we are testing RC's like now it's not too bad to have this information, but let us crop it two two digits from the left (with Boost.PP?), public releases until now did easily more than 10 builds in between.

> I guess if `__CUDACC_VER__` and `__CUDACC_VER_MAJOR__` were introduced at the same time that branch may be redundant anyway.

I think they were.

jzmaddock:
I've updated the patch to allow the build number to go up to 9999, there should still be room for the major version to go up to 1000 without overflow.  I think that should provide us with enough headroom for a while at least?

ax3l:
I think that's enough safety margin ;)

include/boost/config/compiler/nvcc.hpp
```diff
@@ -19,11 +26,11 @@
 // https://svn.boost.org/trac/boost/ticket/11897
 // This is fixed in 7.5. As the following version macro was introduced in 7.5 an existance
 // check is enough to detect versions < 7.5
-#if !defined(__CUDACC_VER__) || (__CUDACC_VER__ < 70500)
+#if BOOST_CUDA_VERSION < 7050000
```
ax3l: Ihkn `7.5` would be `7005000` now, no?

include/boost/config/compiler/nvcc.hpp
```diff
@@ -19,11 +26,11 @@
...
 #   define BOOST_NO_CXX11_VARIADIC_TEMPLATES
 #endif
 // The same bug is back again in 8.0:
-#if (__CUDACC_VER__ > 80000) && (__CUDACC_VER__ < 80100)
+#if (BOOST_CUDA_VERSION > 8000000) && (BOOST_CUDA_VERSION < 8010000)
```
ax3l: I think this should be `BOOST_CUDA_VERSION > 8000000` and `< 9000000`

include/boost/config/compiler/nvcc.hpp
```diff
@@ -43,7 +50,7 @@
 // And this one effects the NVCC front end,
 // See https://svn.boost.org/trac/boost/ticket/13049
 //
-#if (__CUDACC_VER__ >= 80000) && (__CUDACC_VER__ < 80100)
+#if (BOOST_CUDA_VERSION >= 8000000) && (BOOST_CUDA_VERSION < 8010000)
```
ax3l: same here

include/boost/config/compiler/nvcc.hpp
ax3l: ah now, did not see it properly. It's fine

BenjaminW3:
@jzmaddock @ax3l Could this fix please be added to the upcoming boost-1.65.1 release see [here](http://boost.2283326.n4.nabble.com/release-Starting-1-65-1-td4698309.html)?

cdeterman:
Has this been added to the current boost version or part of boost 1.66?

ax3l:
Yes, this is part of boost 1.65.1 and 1.66.0

PR #178, opened 2017-08-17
Improve outdated configuration / new compiler messages.

jzmaddock:


include/boost/config/compiler/sunpro_cc.hpp
```diff
@@ -203,8 +203,8 @@
 #endif
 //
 // last known and checked version is 0x590:
```
jhunold: The comment says the last supported version is "0x590" whereas the new code adjusts this to 0x5150. Maybe just omit the version from the comment or delete it?

jzmaddock:
> The comment says the last supported version is "0x590" whereas the new 
> code adjusts this to 0x5150. Maybe just omit the version from the 
> comment or delete it?
>

IMO the code is self explanatory, so I've removed the version number 
from the comment.

---
http://www.avg.com

PR #179, opened 2017-08-18
MSVC: 14.11 supports structured bindings in C++17 mode

MarcelRaad:
Both config tests pass with /std:c++14 as well as /std:c++17.

### DIFF
--- a/include/boost/config/compiler/visualc.hpp
+++ b/include/boost/config/compiler/visualc.hpp
@@ -196,6 +196,12 @@
 #  define BOOST_NO_CXX14_AGGREGATE_NSDMI
 #endif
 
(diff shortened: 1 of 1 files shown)


include/boost/config/compiler/visualc.hpp
```diff
@@ -196,6 +196,12 @@
 #  define BOOST_NO_CXX14_AGGREGATE_NSDMI
 #endif
 
+// C++17 features supported by VC++ 14.1 (Visual Studio 2017) Update 3
+//
+#if (_MSC_VER < 1911) || (_MSVC_LANG < 201703)
```
KindDragon: But _MSVC_LANG undefined for VS < 2015

MarcelRaad:
Thanks!

PR #182, opened 2017-08-22
Update BOOST_CONSTEXPR_OR_CONST for C++14 where constexpr does not im…

jzmaddock:
…ply const


jzmaddock:
This isn't necessary after all!

PR #191, opened 2017-11-06
Print SD-6 macros and a few more VC STL ones

pdimov:


pdimov:
@jzmaddock What do you think about this?

PR #192, opened 2017-11-17
Typo in visualc.hpp

OlafvdSpek:


jzmaddock:
Thanks for spotting that!

PR #195, opened 2017-12-04
Add BOOST_PRAGMA_MESSAGE

pdimov:
This PR adds `BOOST_PRAGMA_MESSAGE(M)`, a macro that expands to an equivalent on `#pragma message(M)`, when it's supported. In addition, on MSVC it adds `__FILE__` and `__LINE__` to the message, as the plain `#pragma message` has the problem that it's impossible to tell where the message originated.

Since there are places where Boost.Config itself emits messages using the preprocessor, the best place for this utility macro is in Config so that it can be used inside.

This PR includes #194. Also includes a few unrelated doc fixes and regenerates the HTML.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -6,16 +6,11 @@ language: cpp
 
 sudo: false
 
-python: "2.7"
-
-os:
-  - linux
-  - osx
-
 branches:
   only:
     - master
     - develop
+    - /feature\/.*/
 
 env:
   matrix:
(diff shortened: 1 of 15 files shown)


pdimov:
John, would you please consider merging this to master?

pdimov:
In tests, I'm using

```
#if defined(BOOST_NO_CXX11_LAMBDAS)

BOOST_PRAGMA_MESSAGE("Skipping test due to BOOST_NO_CXX11_LAMBDAS")
int main() {}

#else

// ...
```

and I'm going to need to merge such tests to master at some point.

jzmaddock:
Peter: my main machine is out of action at present, so I'm working off an old laptop that has no Boost or even a working compiler till I can figure out the part I need to fix the main one (dodgy lcd cable) :(

Can you take care of this for now please?  Note that everything currently in develop could/should be merged to master really - especially all the msvc-15.5 stuff.

pdimov:
OK, merged develop in https://github.com/boostorg/config/commit/19349c3941adc71a63f67a474a63b3c3f95bb502.

PR #198, opened 2017-12-09
Apply changes required by MSVC-15.5.

jzmaddock:


pdimov:
By default (that is, without `permissive-`) I get failures in `config_test` because of two-phase lookup.

jzmaddock:
I thought /permissive- was the default for all new projects?  The appveyor tests pass.

In any case we need a solution, and I don't see an easy one as there appears to be no way of detecting /permissive-  ?  I do note that nothing in Boost is using BOOST_NO_TWO_PHASE_NAME_LOOKUP at present so maybe this is moot?

pdimov:
New IDE projects are one thing, but in this way you're going to get test failures in the matrix. Appveyor passes because you test with `/permissive-` there.

OlafvdSpek:
> Config.MSVC-14.5: complete configuration of new compiler features.

15.5?

test/config_info.cpp
```diff
@@ -143,19 +143,49 @@ void print_compiler_macros()
...
+  PRINT_MACRO(__AVX2__);
+  PRINT_MACRO(_CHAR_UNSIGNED);
+  PRINT_MACRO(_CLR_VER);
+  PRINT_MACRO(_CONTROL_FLOW_GUARD);
+  PRINT_MACRO(__cplusplus_cli);
+  PRINT_MACRO(__cplusplus_winrt);
   PRINT_MACRO(_CPPRTTI);
+  PRINT_MACRO(_CPPUNWIND);
   PRINT_MACRO(_DLL);
+  PRINT_MACRO(_ISO_VOLATIL);
```
pdimov: Typo? Shouldn't this be `_IS_VOLATILE`?

jzmaddock:
Just out of curiosity, did my replies to your comments (thanks!) come though, or has Github junked them as I don't see them here?

OlafvdSpek:
Mine? I saw commits but no replies I think.

OlafvdSpek:
BTW, is this targeted for 1.66.0, 1.66.1 or 1.67?

jzmaddock:
>Mine? I saw commits but no replies I think.

OK, not getting through then :(

>BTW, is this targeted for 1.66.0, 1.66.1 or 1.67?

Whatever comes after 1.66...

s-martin:
As Microsoft seems to release new compiler versions in much shorter intervals now, are there any ideas yet, how this can be covered by Boost releases?

Are there any workarounds documented for Boost users to overcome the "Unknown compiler" messages, while there is no Boost release, which covers this?

danieljames:
It should say in the warning message, "Define BOOST_CONFIG_SUPPRESS_OUTDATED_MESSAGE to suppress this message".

s-martin:
@danieljames it doesn't, but thanks for the info. I'lll check it out.

danieljames:
Sorry, that message is added in the new release, coming soon hopefully.

PR #209, opened 2018-01-19
Add file/line to BOOST_PRAGMA_MESSAGE on Intel C++

pdimov:


pdimov:
Intel C++ for Linux defines `__GNUC__`. I could test for `__INTEL_COMPILER` in both, I suppose. A matter of taste.

eldiener:
I think it is better to test for Intel twice than to duplicate code but I understand the reason for your choice.

PR #212, opened 2018-02-09
Variable templates are somewhat unusable in GCC 5.1 (bug 65719)

glenfe:
The are only two uses of BOOST_NO_CXX14_VARIABLE_TEMPLATES in Boost right now: Boost.Align (for alignment_of only) and Boost.TypeTraits (for the detection idiom toolkit only), so this should have minimal impact.


jzmaddock:
Sorry for being dumb, but what have variadic templates got to do with that bug?

glenfe:
John, variable-templates, not variadic-templates.

jzmaddock:
>John, variable templates, not variadic templates

Well I did say that I was probably being dumb!  Merging now...

PR #213, opened 2018-02-24
BOOST_STATIC_CONSTANT should use constexpr

Kojoley:


Lastique:
This broke at least Boost.MPL and all dependent libraries:

```
gcc.compile.c++ bin.v2/libs/thread/build/gcc-4.7/debug/threadapi-pthread/threading-multi/pthread/once.o
In file included from ./boost/thread/detail/invoke.hpp:32:0,
                 from ./boost/thread/pthread/once_atomic.hpp:17,
                 from ./boost/thread/once.hpp:26,
                 from libs/thread/src/pthread/./once_atomic.cpp:9,
                 from libs/thread/src/pthread/once.cpp:8:
./boost/mpl/bool.hpp:34:23: error: redeclaration ‘mpl_::bool_<B>::value’ differs in ‘constexpr’
./boost/mpl/bool.hpp:25:5: error: from previous declaration ‘value’
./boost/mpl/bool.hpp:34:23: error: declaration of ‘constexpr const bool value’ outside of class is not definition [-fpermissive]
    "g++-4.7"   -std=c++11 -fPIC -m64 -pthread -O0 -fno-inline -Wall -pedantic -g -Wextra -Wno-long-long -Wno-unused-parameter -Wunused-function -pedantic -DBOOST_ALL_NO_LIB=1 -DBOOST_SYSTEM_DYN_LINK=1 -DBOOST_THREAD_BUILD_DLL=1 -DBOOST_THREAD_DONT_USE_CHRONO -DBOOST_THREAD_POSIX  -I"." -c -o "bin.v2/libs/thread/build/gcc-4.7/debug/threadapi-pthread/threading-multi/pthread/once.o" "libs/thread/src/pthread/once.cpp"
```

jzmaddock:
Thanks for the heads up - I've reverted the change - if someone wants to re-submit a new PR that fixes the problem that's fine, otherwise I think I'll leave well alone...

jzmaddock:
BTW, is there any use case where constexpr delivers something that the good old `static const T = ` does not?  I'm struggling to think of one.

Kojoley:
It guarantees that your constant is computable at compile time so you will get error message on definition rather then on usage.

jzmaddock:
>It guarantees that your constant is computable at compile time so you will get error message on definition rather then on usage.

Assuming you're using BOOST_STATIC_CONST at class scope, which is really the intended usage, that's true in any case surely?  There may be exceptions for literal classes - but since these are C++11 or later anyway, you might as well just be using constexpr directly.

There are a lot of corner cases - for example from C++17 we could add the `inline` declaration so that an out-of-line definition isn't required and so on...

Lastique:
I think we should deprecate `BOOST_STATIC_CONST` in favor of `static BOOST_CONSTEXPR_OR_CONST`. That way at least the declaration and definition can be made consistent more easily. Enum workaround is no longer relevant, I think.

Regarding C++17 inline variables, we need `BOOST_CXX17_INLINE_VARIABLE` (I have one in Boost.Atomic already), but that is another PR.

Kojoley:
> Assuming you're using BOOST_STATIC_CONST at class scope, which is really the intended usage, that's true in any case surely? There may be exceptions for literal classes

I do not see any.

> If a static data member is of const literal type, its declaration in the class definition can specify a brace-orequal-initializer in which every initializer-clause that is an assignment-expression is a constant expression.
> A static data member of literal type can be declared in the class definition with the constexpr specifier; if so, its declaration shall specify a brace-or-equal-initializer in which every initializer-clause that is an assignment-expression is a constant expression. [ Note: In both these cases, the member may appear in constant expressions. — end note ] The member shall still be defined in a namespace scope if it is odrused (3.2) in the program and the namespace scope definition shall not contain an initializer.

http://www.open-std.org/jtc1/sc22/wg21/docs/papers/2011/n3242.pdf § 9.4.2/3

```cpp
struct Foo {
  int i;
  int j;
};

struct Bar {
  static constexpr Foo foo = /*Foo*/{};
};
```

Lastique:
@Kojoley This is only possible in C++11 and later. C++03 [class.static.data]/4 only allows in-class initializer for const integral or const enumeration types. If you're writing C++11 code, you don't need `BOOST_STATIC_CONSTANT`.

Lastique:
I think his point was that replacing `const` with `constexpr` in the context of `BOOST_STATIC_CONSTANT` does not offer any benefits. And it doesn't, except to make the code look more modern.

Kojoley:
It looks like you are right. I though this would be caught at compile time, but it does not:
```cpp
struct A 
{
    static constexpr int value = 123;
    void foo(const int &) { }
    void bar() { foo(value); }
};

int main()
{
    A a;
    a.bar();
}
```

jzmaddock:
>I think his point was that replacing const with constexpr in the context of BOOST_STATIC_CONSTANT does not offer any benefits. And it doesn't, except to make the code look more modern.

Yes that was exactly my point.  And if it's all hidden behind a macro then it doesn't really look very modern either :(

A similar situation exists with `inline` declared class variables:

* If you're targeting C++17 and later just go ahead and use them.
* If you're targeting earlier versions, then an out-of-line definition will be required in some use contexts, so you might as well forget the "inline" for C++17 and just go ahead and provide the definition.  If you start macro-izing the code with different branches for pre and post C++17, then you end up with more typing not less - which rather defeats the whole point of the feature.

Lastique:
C++17 inline variables can be used not only as static class data members. In Boost.Atomic, I'm using them to declare namespace-scope constants. The code compiles whether these constants are declared inline or not, but there's a subtle difference in linkage, which may be significant in some cases.

https://github.com/boostorg/atomic/blob/develop/include/boost/memory_order.hpp#L57

pdimov:
constexpr is inline anyway.

In C++17, the out-of-class definition of the constexpr member variable is deprecated, too. It's neverending fun.

Lastique:
> constexpr is inline anyway.

Is it? I haven't found where it is said so.

pdimov:
"A function or static data member declared with the constexpr specifier is implicitly an inline function or variable ([dcl.inline])."

http://eel.is/c++draft/dcl.constexpr#1

Lastique:
Oh, you meant a static data member. I thought you meant any constexpr variables.

pdimov:
I think I remember suggestions to make all namespace-scope or static constexpr inline (makes sense to me), but apparently the CWG decided against that.

Here's the deprecation reference: http://eel.is/c++draft/depr.static_constexpr

PR #214, opened 2018-03-06
Add BOOST_NO_CXX17_IF_CONSTEXPR.

tzlaine:


tzlaine:
This is needed for integrating Yap, which is in in post-review make ready.

tzlaine:
Looks like there are MSVC problems.  I'll try again.

PR #216, opened 2018-03-07
Add BOOST_NO_CXX17_IF_CONSTEXPR.

tzlaine:


tzlaine:
Third time's the charm.  This passes all tests now.  Needed by Yap before it can be integrated as a Boost module.

Lastique:
It might be useful to verify in the test the `if constexpr` statement actually discards one of the branches. For example, add a failing static assert in a dependent context.

jzmaddock:
Other than Lastique's useful suggestion, this all looks good to me.  However, I won't merge until after the release.

tzlaine:
Good idea.  I've added a static_assert to one of the false paths.

PR #223, opened 2018-04-15
Bump to 1.68.0

danieljames:


morinmorin:
I think we're ready for version bump, so let's merge it now.

PR #224, opened 2018-04-28
Add parentheses to BOOST_CUDA_VERSION

pdimov:
Expressions like `BOOST_CUDA_VERSION / 1000000` don't work correctly; add parentheses around the expression to fix.


yuyi1005:
I use vs2017.4+cuda9.1+boost1.67
When including a boost header file in .cu file, it cannot compile. 

fatal error C1012: parentheses not match

in file "boost/preprocessor/slot/detail/shared.hpp"
line 27:
 # if (BOOST_PP_VALUE) / 1000000000UL == 0

Do you have any idea to cope with this issue? Thanks

PR #228, opened 2018-07-12
Prefix "Unknown compiler version ..." messages with "boost:".

lanurmi:
Makes it easier to figure out which library is outputing such messages
during compilation.

Especially useful with MSVC.


lanurmi:
Eh, even though I forked the repo *today*, and made my changes, looks like I got a revision from 2016, and since then the most important changes have already been made in develop, in a slightly better way.

I leave it up to you to decide whether the rest are worth merging in their current form.

PR #242, opened 2018-09-14
[CMake] Generate cmake target for boost config that other libraries can depend on

Mike-Devel:
For further reference: https://groups.google.com/forum/#!topic/boost-developers-archive/kM4JRmyVl3M

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -264,7 +264,10 @@ install:
   - cp -r $TRAVIS_BUILD_DIR/* libs/config
   - ./bootstrap.sh
   - ./b2 headers
- 
+  - cd ..
+  - git clone -b $TRAVIS_BRANCH --depth 1 https://github.com/boostorg/config.git boost-config-standalone
+  - mkdir cmake-build
+
 script:
   - |-
     echo "using $TOOLSET : : $COMPILER ;" > ~/user-config.jam
@@ -272,6 +275,9 @@ script:
   - for CXXLOCAL in $CXXSTD; do  (cd libs/config/test && ../../../b2 config_info_travis_install toolset=$TOOLSET cxxstd=$CXXLOCAL $CXXSTD_DIALECT && echo With Standard Version $CXXLOCAL && ./config_info_travis && rm ./config_info_travis)  done
   - unset IFS
   - ./b2 -j3 libs/config/test toolset=$TOOLSET cxxstd=$CXXSTD $CXXSTD_DIALECT
+  - cd ../cmake-build
+  - cmake ../boost-config-standalone/test/cmake
+  - cmake --build .
 
 notifications:
   email:
--- a/.travis.yml
+++ b/.travis.yml
@@ -264,10 +264,7 @@ install:
   - cp -r $TRAVIS_BUILD_DIR/* libs/config
   - ./bootstrap.sh
   - ./b2 headers
-  - cd ..
-  - git clone -b $TRAVIS_BRANCH --depth 1 https://github.com/boostorg/config.git boost-config-standalone
-  - mkdir cmake-build
-
+ 
 script:
   - |-
     echo "using $TOOLSET : : $COMPILER ;" > ~/user-config.jam
@@ -275,9 +272,6 @@ script:
   - for CXXLOCAL in $CXXSTD; do  (cd libs/config/test && ../../../b2 config_info_travis_install toolset=$TOOLSET cxxstd=$CXXLOCAL $CXXSTD_DIALECT && echo With Standard Version $CXXLOCAL && ./config_info_travis && rm ./config_info_travis)  done
   - unset IFS
   - ./b2 -j3 libs/config/test toolset=$TOOLSET cxxstd=$CXXSTD $CXXSTD_DIALECT
-  - cd ../cmake-build
-  - cmake ../boost-config-standalone/test/cmake
-  - cmake --build .
 
 notifications:
   email:
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -1,3 +1,7 @@
+# Copyright 2018 Mike Dev
+# Distributed under the Boost Software License, Version 1.0.
+# See accompanying file LICENSE_1_0.txt or copy at http://www.boost.org/LICENSE_1_0.txt
+
 cmake_minimum_required(VERSION 3.5)
 project(BoostConfig LANGUAGES CXX)
 
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -1,0 +1,7 @@
+cmake_minimum_required(VERSION 3.8)
+project(boost-config LANGUAGES CXX)
+
+add_library(boost_config INTERFACE)
+add_library(Boost::config ALIAS boost_config)
+
+target_include_directories(boost_config INTERFACE include)
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -1,5 +1,5 @@
-cmake_minimum_required(VERSION 3.8)
-project(boost-config LANGUAGES CXX)
+cmake_minimum_required(VERSION 3.5)
+project(BoostConfig LANGUAGES CXX)
 
 add_library(boost_config INTERFACE)
 add_library(Boost::config ALIAS boost_config)
--- a/test/cmake/CMakeLists.txt
+++ b/test/cmake/CMakeLists.txt
@@ -1,0 +1,13 @@
+# Copyright 2018 Mike Dev
+# Distributed under the Boost Software License, Version 1.0.
+# See accompanying file LICENSE_1_0.txt or copy at http://www.boost.org/LICENSE_1_0.txt
+
+cmake_minimum_required(VERSION 3.5)
+project(BoostFilesystem_test)
+
+add_subdirectory(../.. ${CMAKE_CURRENT_BINARY_DIR}/boost_config)
+
+add_executable(boost_config_cmake_compilation_test main.cpp)
+target_link_libraries(boost_config_cmake_compilation_test Boost::config)
+
+
--- a/test/cmake/main.cpp
+++ b/test/cmake/main.cpp
@@ -1,0 +1,7 @@
+// Copyright 2018 Mike Dev
+// Distributed under the Boost Software License, Version 1.0.
+// See accompanying file LICENSE_1_0.txt or copy at http ://www.boost.org/LICENSE_1_0.txt
+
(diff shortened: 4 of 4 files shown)


pdimov:
Would be nice if this (with an appropriate copyright banner added) could get merged, because without it, CMakeLists.txt in other libraries are unusable as they all refer to Boost::config.

For instance, I've added, as an experiment, CMake-based testing to Assert, but it presently fails:

https://travis-ci.org/boostorg/assert/jobs/430784534

jzmaddock:
Apologies, I'm way behind :(

Question: lets say we merge this and folks folks start depending on it "because it's there", do we have an issue later if we adopt a different boost-wide solution later?

eldiener:
We also do not currently have any regression or CI way to test these CMake additions to show that they actually work in CMake based projects which use Boost. Boost policy has always been that there must be a way to verify that changes work correctly, not just that they look correct.

Mike-Devel:
@jzmaddock:
> do we have an issue later if we adopt a different boost-wide solution later?

Nothing that I can think of as there is (deliberately) so little functionality in it.
 
One thing I could imagine is that whatever solution comes up later chooses a different naming scheme (e.g. `boost::config` vs `Boost::config`). I would be trivial to support both, so maintaining backwards compatibility would not be a problem, but of course it is not an ideal situation.

Mike-Devel:
@eldiener: I can absolutely add a test that gets run by travis. I probably don't know enough about the rest of the boost test infrastructure to integrate it there.

Mike-Devel:
Looking at the travis file, the main difficulty is actually to isolate the cmake treaty against the b2 stuff and vice versa, to make sure it really works without it.

pdimov:
It's possible to work on testing via CMake (one way or another), but without Config, nothing can proceed.

https://github.com/boostorg/assert/blob/develop/CMakeLists.txt has support for testing and https://github.com/boostorg/assert/blob/feature/cmake-test/.travis.yml#L37 uses it, but it fails without Config having CMakeLists.

pdimov:
It's better to create a separate Travis job with its own install/script entries instead of attaching the CMake tests everywhere.

Mike-Devel:
OK, I added a test for the cmake script. As I can't test the travis script localy it might require a few iterations.

pdimov:
Attaching the Travis changes to this PR lowers its chances of being accepted IMO.

Mike-Devel:
@pdimov: Thats one of the reasons I didn't do it originally. Much more things to bikeshed about ;). If jzmaddock doesn't like it, I have no problem with removing it again, but at least people can see that "it works".

pdimov:
At least do it right then.

```
- os: linux
  install: true
  script:
    - mkdir __build__ && cd __build__
    - cmake ../test/cmake
    - cmake --build .
```

:-)

pdimov:
Either way, I think that Config is not the right place to experiment with testing CMake files. We'll do that elsewhere and only add it here when it's ready. We just need the basic CMakeLists here because Config is a dependency of everything and the elsewhere work can't proceed.

Mike-Devel:
Fair enough.  I'll only be able to revert this tomorrow though.

MikeGitb:
Sorry, life happened ;). 
Should be able to do this by the end of the day, but if anyone is eager to merge it it isn't like you can't merge and then revert the last commit.

Mike-Devel:
Sorry for the delay - life happend

jzmaddock:
Still not sure we're doing the right thing... but now merged.  Enjoy!

jzmaddock:
I've pushed some comments to the CMake files to the effect that these are experimental.

Some remaining issues:

* I think that test/cmake/* should all go, in particular it contains `project(BoostFilesystem_test)` that looks like a flat out bug?
* If we're adding tests, please add a cmake file directly to /test/ that builds config_info.cpp and config_test.cpp as separate tests.

glenfe:
> I think that test/cmake/* should all go

Yes. The only thing needed here is the one CMakeLists.txt. Feel free to nuke whatever that test/cmake is.

Mike-Devel:
@jzmaddock: Thanks 

> I think that test/cmake/* should all go, in particular it contains project(BoostFilesystem_test) that looks like a flat out bug?

That folder contains the testproject for the `CMakeLists.txt` file requrested by @eldiener (i.e. can the cmake file used by another cmake based project). So it's purpose is  not to run the Boost.config tests but to test the cmake file itself. 
And yes, the `project(BoostFilesystem_test)` is a typo (although an inconsequential one) and should have been `project(BoostConfig_test)`.I originally had that test run by travis-CI, but removed that as requested by @pdimov. It can still be usefull for manual testing though.

If you don't think such a test is necessary, then just remove that folder.

jzmaddock:
> Feel free to nuke whatever that test/cmake is.

Nuclear option exercised :)

If someone wants to submit a CMake testing script that's fine, but it should actually build and run the existing tests...

Mike-Devel:
> If someone wants to submit a CMake testing script that's fine, but it should actually build and run the existing tests...

Again, that was not the purpose of the script. I just did what @eldiener requested

Mike-Devel:
Btw.: Does anyone have an opinion on the proposed top level cmake file (https://github.com/boostorg/boost/pull/193)?

eldiener:
The only thing I have requested, and it appears to have been possibly misunderstood, is that if a CMake PR is done for a particular library, then that specific library should have a means to test that the CMake PR is correct and does what it is supposedf to be doing, which I gather is to make the particular library usable by end-user CMake scripts. I have not requested that CMake be used in place of b2 to run tests in a particular library's test directory, if that is what is actually believed. I believe that a CMake PR should be treated like any other PR; there should be a way to test that it is correct for the particular library, or else what is the point of the PR.

pdimov:
That was actually what the addition did (tried to do) - it added a mock user project in `test/cmake` (which could probably have been better named as `test/cmake_test_project`) and then built it. The Travis addition was however wrong.

eldiener:
The 'test/cmake' is a directory in each particular library ? How would the preprocessor library ( or any library ) verify that the proposed CMake PR for that library is correct ? The usual way of doing this is to add a test to the bjam file in the library's test subdirectory which verifies that the PR is valid, if some test does not already exist which tests the validity of the PR. Is it not possible to do this with the CMake PRs being proposed ?

pdimov:
If you want `cmake_test.cpp` in the `test` directory, which is then executed by `b2 test`, to test the CMake file, that's going to be a bit more convoluted because the functionality isn't in C++. The test will basically need to execute the `cmake` commands.

I see that Mike has posted a question to that effect on the list.

It's easier to test the sequence of `cmake` commands in a Travis job instead of from a C++ test program, but the other way is I suppose possible too.

I've re-added the CMake test project to a feature branch here: https://github.com/boostorg/config/commit/bef8ef7b80814d58680522dd898c0aeffbef819e

We'll need to wait on Travis a bit to see if I managed to get it right.

pdimov:
Yes I did: https://travis-ci.org/boostorg/config/jobs/434773579

Mike-Devel:
@eldiener:
> , then that specific library should have a means to test that the CMake PR is correct and does what it is supposedf to be doing, which I gather is to make the particular library usable by end-user CMake scripts.

Which is exactly what that (first nuked and now fixed & resurected) code did.

> I have not requested that CMake be used in place of b2 to run tests in a particular library's test directory, if that is what is actually believed.

No worries, I never understood it that way, nor did I ever had the plan to provide cmake files that run the regular library unit tests.

@pdimov:
> I've re-added the CMake test project to a feature branch here: bef8ef7

Thanks. Maybe you should also a job for Mac? IT is not really relevant here

@glenfe:
> Feel free to nuke whatever that test/cmake is.

Sorry if that sounds childish, but do you always recommend to delete other people's code whose purpose you don't understand? I mean, just like you I'm only interested in the top level CMakeLists.txt file, so I don't really care what happens to those particular files, but generally that seems not like an ideal approach.

glenfe:
> but do you always recommend to delete other people's code

Not always. In circumstances where the particulars are a facsimile of the ones in this circumstance, always.

> I mean, just like you I'm only interested in the top level CMakeLists.txt file, so I don't really care what happens to those particular files

Excellent. Then they won't be missed by either of us.

glenfe:
To be clear, the maintainer of the library here was rightly concerned by whatever you had attempted to do in test/cmake because it:

* Looked clearly wrong (should not have even involved Boost.Filesystem)
* More importantly, did not even work (as indicated by the Travis results)

That he raised concerns about the above, that he inquired about the removal of those files, that he was encouraged to remove them, that he did remove them with the commit message "Remove malformed CMake testing files", or that he subsequently posted "If someone wants to submit a CMake testing script that's fine, but it should actually build and run the existing tests..." shouldn't surprise anyone.

jzmaddock:
Before I start adding more stuff from other PR's, just a quick question about naming conventions - be prepared for a full bike shed discussion, and no this probably isn't the right forum...but - this adds 2 named targets for the library - "boost_config" and "Boost::config".  Now, normally we tend to use "Boost.LibraryName" when referring to a particular library, and if we're using a "::" separator, then we should follow the namespace name, so all lowercase "boost::library".  Thoughts?

pdimov:
We can go either way on that one and I originally did use `boost::library`. However, capital case for the target namespace is the CMake convention, and the existing FindBoost.cmake naming does use f.ex. `Boost::filesystem` for the targets. So... `Boost::` it is.

Mike-Devel:
A bit late but anyhow: 
- `boost_config should be considered a private variable name. No one should reference it and every library maintainer may change the name at their leisure . Unfortunately, I don't thinkg cmake provides an easy way to make targets only referencable locally (Niall Dougles mentioned one way on the boost ML). 

- `Boost::config` - as explained by pdimov - follows the naming convention set by cmake itself (e.g. if cmake can find a system installed Boost.Filesystem library it will provide a `Boost::filesystem` target that can be imported via `find_package(Boost COMPONENTS filesystem)` (https://cmake.org/cmake/help/v3.13/module/FindBoost.html). 
Obviously there are advantages and disadvantages to this: Danger of collision / mixup vs consistency. One reason I didn't consider the collison a dealbreaker is that :
  a) As far as I can tell, cmake will always prefer the local target, which is what we want
  b) Even if the boost targets are named differently to prevent collisions on the cmake level, your compiler would still find the system installed boost headers and libraries anyway, so I don't know a reliable way to prevent mixups anyway (which is one reason I don't like system package managers used for dependency resolution during c++ development).

Issue #247, opened 2018-10-16
CMake feature detection


rleigh-codelibre:
As mentioned in https://lists.boost.org/Archives/boost/2018/10/243729.php I would be happy to implement this type of functionality.

Would a PR implementing the detection logic be acceptable?  Or is someone already working on this?

Thanks,
Roger

jzmaddock:
I'm quite certain that I'm not qualified to comment - anyone else?

eldiener:
I am not qualified to judge a CMake implementation either. I did suggest that doing this for config ( and predef also ) and then having current Boost libraries which are converted to CMake use the CMake implementations in config and predef to replace the bjam code they already have now is a much better solution than having each library do their own CMake implementation for what they will need.

rleigh-codelibre:
I have done some initial work on this in the linked PR (#249).  There will be more to come, this is just the initial step to make it installable and re-usable by downstream consumers.

PR #249, opened 2018-10-25
cmake: Install headers and exported configuration

rleigh-codelibre:
Initial work for https://github.com/boostorg/config/issues/247 which extends the existing two cmake files.  This makes it possible to install the headers with `make install` and it also builds two tests.  It additionally installs a configuration file so that you can run e.g.

```
find_package(BoostConfig VERSION 1.69 REQUIRED)
```

and then be able to use

```
target_link_libraries(foo Boost::config)
```

in any downstream project making use of Boost.Config.  The generated configuration is put into a "BoostConfigInternal" script, which is then sourced by a minimal "BoostConfig" script.  This is so that we can set additional variables for build-time configuration in this top-level script as follow-on work.

Quite a few bits of the logic here could be generalised and reused by other Boost components as done by BCM.  However, for now I have kept it dead simple.  It's easy enough to refactor it later on to be more complex.  This is the minimal work needed to make it possible to use Boost::config in other projects.

Hope this is useful,
Roger

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -16,12 +16,6 @@
 cmake_minimum_required(VERSION 3.5)
 project(BoostConfig VERSION 1.69 LANGUAGES CXX)
 
-# Globbing is not recommended, but it matches the b2 behaviour for
-# now.  Should be a static list.
-file(GLOB_RECURSE boost_config_headers
-  RELATIVE "${CMAKE_CURRENT_SOURCE_DIR}/include"
-  CONFIGURE_DEPENDS "${CMAKE_CURRENT_SOURCE_DIR}/include/boost/*")
-
 add_library(boost_config INTERFACE)
 add_library(Boost::config ALIAS boost_config)
 
@@ -33,13 +27,9 @@ set_target_properties(boost_config PROPERTIES EXPORT_NAME "Boost::config")
 
 # Install headers
 include(GNUInstallDirs)
-foreach(header ${boost_config_headers})
-  get_filename_component(header_name "${header}" NAME)
-  get_filename_component(header_dir "${header}" DIRECTORY)
-  install(FILES "include/${header}"
-    DESTINATION "${CMAKE_INSTALL_INCLUDEDIR}/${header_dir}"
-    COMPONENT "development")
-endforeach()
+install(DIRECTORY "${CMAKE_CURRENT_SOURCE_DIR}/include/boost"
+  DESTINATION "${CMAKE_INSTALL_INCLUDEDIR}"
+  COMPONENT "development")
 
 # Install exported configuration
 if(WIN32)
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -1,4 +1,5 @@
 # Copyright 2018 Mike Dev
+# Copyright 2018 Roger Leigh
 # Distributed under the Boost Software License, Version 1.0.
 # See accompanying file LICENSE_1_0.txt or copy at http://www.boost.org/LICENSE_1_0.txt
 #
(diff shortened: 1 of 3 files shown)


rleigh-codelibre:
@Mike-Devel This might be of interest to you if you wanted to review it.

test/CMakeLists.txt
```diff
@@ -16,12 +17,12 @@
...
 cmake_minimum_required(VERSION 3.5)
-  
-add_subdirectory(.. ${CMAKE_CURRENT_BINARY_DIR}/boost_config)
-  
-add_executable(config_info config_info.cpp)  
-target_link_libraries(config_info Boost::config)  
 
-add_executable(config_test config_test.cpp)  
-target_link_libraries(config_test Boost::config)  
+find_package(Threads)
```
Mike-Devel: Why does this cmake self test have to enable multithreading?

CMakeLists.txt
```diff
@@ -13,9 +14,62 @@
...
+write_basic_package_version_file(
+  ${CMAKE_CURRENT_BINARY_DIR}/${PROJECT_NAME}ConfigVersion.cmake
+  VERSION "${PROJECT_VERSION}"
+  COMPATIBILITY SameMinorVersion)
+install(FILES ${CMAKE_CURRENT_BINARY_DIR}/${PROJECT_NAME}Config.cmake
+              ${CMAKE_CURRENT_BINARY_DIR}/${PROJECT_NAME}ConfigVersion.cmake
+        DESTINATION "${boost_config_config_dir}"
+        COMPONENT "development")
+
+add_subdirectory(test)
```
Mike-Devel: At the moment, the purpose of the CMake file in test is not to run the unit tests, but to check, if the top level cmake file can be used from an external project. As such the original order (<root>/test/CMakeLists.txt calls <root>/CMakeLists.txt) seems more appropriate

CMakeLists.txt
```diff
@@ -13,9 +14,62 @@
...
 
 cmake_minimum_required(VERSION 3.5)
-project(BoostConfig LANGUAGES CXX)
+project(BoostConfig VERSION 1.69 LANGUAGES CXX)
+
+# Globbing is not recommended, but it matches the b2 behaviour for
+# now.  Should be a static list.
+file(GLOB_RECURSE boost_config_headers
+  RELATIVE "${CMAKE_CURRENT_SOURCE_DIR}/include"
+  CONFIGURE_DEPENDS "${CMAKE_CURRENT_SOURCE_DIR}/include/boost/*")
```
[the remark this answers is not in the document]
rleigh-codelibre: It looks like `install(DIRECTORY …)` might work in this case.  Most other projects I've worked on had more complex arrangements where this didn't work.

CMakeLists.txt
```diff
@@ -13,9 +14,62 @@
...
+  set(boost_config_config_dir "cmake")
+else()
+  set(boost_config_config_dir "${CMAKE_INSTALL_LIBDIR}/cmake/${PROJECT_NAME}")
+endif()
+
+install(TARGETS boost_config
+        EXPORT BoostConfigInternal
+        RUNTIME DESTINATION ${CMAKE_INSTALL_BINDIR}
+        LIBRARY DESTINATION ${CMAKE_INSTALL_LIBDIR}
+        ARCHIVE DESTINATION ${CMAKE_INSTALL_LIBDIR}
```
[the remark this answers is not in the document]
rleigh-codelibre: They can probably be omitted for this specific library; they are they solely for completeness if used in other projects.

test/CMakeLists.txt
rleigh-codelibre: The test requires it (fails with pthread error if omitted):

```
[ 25%] Linking CXX executable config_test
/usr/bin/ld: CMakeFiles/config_test.dir/config_test.cpp.o: in function `boost_has_pthread_mutexattr_settype::f()':
config_test.cpp:(.text+0x5ace): undefined reference to `pthread_mutexattr_init'
/usr/bin/ld: config_test.cpp:(.text+0x5ae6): undefined reference to `pthread_mutexattr_settype'
/usr/bin/ld: CMakeFiles/config_test.dir/config_test.cpp.o: in function `boost_has_pthread_yield::f()':
config_test.cpp:(.text+0x5b15): undefined reference to `pthread_yield'
/usr/bin/ld: CMakeFiles/config_test.dir/config_test.cpp.o: in function `boost_has_pthreads::test()':
config_test.cpp:(.text+0x5b93): undefined reference to `pthread_mutex_trylock'
/usr/bin/ld: config_test.cpp:(.text+0x5bd1): undefined reference to `pthread_create'
/usr/bin/ld: config_test.cpp:(.text+0x5bed): undefined reference to `pthread_join'
collect2: error: ld returned 1 exit status
make[2]: *** [test/CMakeFiles/config_test.dir/build.make:84: test/config_test] Error 1
make[1]: *** [CMakeFiles/Makefile2:93: test/CMakeFiles/config_test.dir/all] Error 2
make: *** [Makefile:130: all] Error 2
```

CMakeLists.txt
rleigh-codelibre: OK.  It's rather unusual, but I can revert this if that's how it's intended to be.

rleigh-codelibre:
@Mike-Devel I've pushed some changes which should address the remainder of your comments.

CMakeLists.txt
```diff
@@ -13,9 +14,47 @@
 #
 
 cmake_minimum_required(VERSION 3.5)
-project(BoostConfig LANGUAGES CXX)
+project(BoostConfig VERSION 1.69 LANGUAGES CXX)
```
pdimov: I can pretty much guarantee that this will become incorrect if not tested. In fact it already is.

pdimov: See https://github.com/boostorg/assert/blob/develop/test/CMakeLists.txt#L6 and https://github.com/boostorg/assert/blob/develop/test/check_cmake_version.cpp for one possible approach.

rleigh-codelibre: Longer term, if the components are decoupled and usable independently, the versioning can then diverge between components.

I'm not sure how the links you provided make a solution though.  Do you want to build and run a test program to get the version?  If so, would this not prove problematic for cross compilation?

Would it not be possible to copy the logic here: https://github.com/Kitware/CMake/blob/master/Modules/FindBoost.cmake#L1370 so that we always use the version parsed directly from the headers?

rleigh-codelibre:
I'm not doing work on this going forward.  If anyone wants to pick this up, please feel free to do so.  What's here is out of data and likely no longer useful.

PR #261, opened 2019-02-08
Tentatively enable __float128 support for clang

jzmaddock:
See https://github.com/boostorg/math/issues/181


NAThompson:
Not super knowledgeable about this, but wouldn't it be safer to check for clang_major >= 7? I thought the problem was instruction generation for clang <= 6.

PR #268, opened 2019-03-29
Lift thread local ban for 64bit MinGW

Kojoley:
The problem seems affect only 32bit compilers. See https://sourceforge.net/p/mingw-w64/bugs/527/#cedc.

ref https://github.com/boostorg/config/commit/fe5e07b521e49f6cca712c801e025fed13c23979#commitcomment-32969838


Kojoley:
Ping

PR #274, opened 2019-04-03
Print config info via b2

Kojoley:
Simplifies CI build scripts. Allows having CI builds with multiple options
like `address-model=32,64`.


jzmaddock:
Thanks for that!

jzmaddock:
For some reason this caused a ton of failures when merged.
It's a good idea but reverted for now.

Kojoley:
The build before reverting the PR have passed. I have no idea what was the problem, it is also passing in my fork.

jzmaddock:
Now that everything else has stabilized, I've reapplied the patch in: https://github.com/boostorg/config/pull/290 and we'll see what happens this time.  It worked for me locally as well BTW.

PR #278, opened 2019-04-12
Bump version to 1.71.0

mclow:


pdimov:
Merging to unblock autolink/FindBoost users.

Issue #300, opened 2019-10-19
BOOST_ATTRIBUTE_NODISCARD triggers -Wpedantic on clang-6


HDembinski:
See issue on [Godbolt](https://godbolt.org/#g:!((g:!((g:!((h:codeEditor,i:(j:1,lang:c%2B%2B,source:'%23include+%3Cboost/config.hpp%3E%0A%0Astruct+BOOST_ATTRIBUTE_NODISCARD+Foo+%7B%7D%3B%0A%0Aint+main()+%7B%0A%7D%0A'),l:'5',n:'0',o:'C%2B%2B+source+%231',t:'0')),k:50,l:'4',n:'0',o:'',s:0,t:'0'),(g:!((g:!((h:compiler,i:(compiler:clang600,filters:(b:'0',binary:'1',commentOnly:'0',demangle:'0',directives:'0',execute:'1',intel:'0',libraryCode:'1',trim:'1'),lang:c%2B%2B,libs:!((name:boost,ver:'171')),options:'-O3+-DNDEBUG+-Wall+-Wpedantic+-Werror',source:1),l:'5',n:'0',o:'x86-64+clang+6.0.0+(Editor+%231,+Compiler+%231)+C%2B%2B',t:'0')),k:50,l:'4',m:71.99297629499561,n:'0',o:'',s:0,t:'0'),(g:!((h:output,i:(compiler:1,editor:1,wrap:'1'),l:'5',n:'0',o:'%231+with+x86-64+clang+6.0.0',t:'0')),header:(),l:'4',m:28.007023705004386,n:'0',o:'',s:0,t:'0')),k:50,l:'3',n:'0',o:'',t:'0')),l:'2',n:'0',o:'',t:'0')),version:4)

I am not quite sure what to do about this, there are several options. I think this is not our fault, clang is inconsistent here. My expectation is that when I compile with `-std=c++14`, a compiler should simply not accept the attribute. But clang-6 does accept it anyway which is why the configuration tests pick this up. However, when pedantic warnings are enabled, it warns about the usage.

So the general solution would be to run the configuration tests with `-Wpedantic -Werror`.

HDembinski:
After looking into the code, I noticed that there is no configuration test for [[nodiscard]]. So this can be fixed in a simpler way. I will make a PR.

jzmaddock:
Fixed in develop

HDembinski:
Cool, thanks!

PR #301, opened 2019-10-19
fixes warning in clang-6

HDembinski:
Addresses #300

### DIFF
--- a/include/boost/config/detail/suffix.hpp
+++ b/include/boost/config/detail/suffix.hpp
@@ -996,7 +996,8 @@ namespace std{ using ::type_info; }
 // [[nodiscard]]:
 //
 #ifdef __has_cpp_attribute
-#if __has_cpp_attribute(nodiscard)
+// clang-6 accepts [[nodiscard]] with -std=c++14, but warns about it -pedantic
+#if __has_cpp_attribute(nodiscard) && !(defined(__clang__) && (__cplusplus < 201703L))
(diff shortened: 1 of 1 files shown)


jzmaddock:
Thanks for that.

Can you please report upstream to clang, as it seems unreasonable for `__has_cpp_attribute(nodiscard)` to report `true` when `-pedantic -Werror` will result in a compiler error!

HDembinski:
I will try to report it upstream. Registering on the LLVM bug-tracker is surprisingly complicated, apparently they are having problems with spam.

HDembinski:
Actually, maybe we misunderstood the usage of `__has_cpp_attribute`. Despite the name, it does not only return true and false, but the date when the feature was added to the C++ standard.
https://en.cppreference.com/w/cpp/feature_test

So we maybe we should test like this?

```
#if __has_cpp_attribute(nodiscard)  >= __cplusplus
#define BOOST_ATTRIBUTE_NODISCARD [[nodiscard]]
#else
#define BOOST_ATTRIBUTE_NODISCARD
#endif
```

jzmaddock:
The problem is that msvc still defines __cplusplus to 199711 :(

HDembinski:
Ah, ok. Then we should probably keep it as is. :)

jzmaddock:
>Ah, ok. Then we should probably keep it as is. :)

That's the easy solution, but I suspect that technically you are correct and this may all come back to bite us whatever....

PR #307, opened 2019-12-02
Do not inject `typeinfo` into `std` namespace

### LINKED ISSUE #306: `<typeinfo>` unconditionally included on dinkumware
It was added in 668b3fccaef7460036a1a83a9c90df8cbe76c4ab to fix lexical_cast (https://svn.boost.org/trac10/ticket/4115) problem, but that seems to be an overkill (and probably not needed since https://github.com/boostorg/lexical_cast/commit/3ce36a28484b76fab2dd96e8ee43f52ae87d1353?).
Also, it is a strange thing that with defined `BOOST_NO_STD_TYPEINFO` there is `std::typeinfo`. It completely undermines the purpose of the macro and of `boost/core/typeinfo.hpp`.

Kojoley:
`BOOST_NO_STD_TYPEINFO` should be used to detect if there is no `typeinfo` in `std` namespace, or Boost.Core/Boost.TypeInfo wrappers for a portable solution.

Closes #306.

### DIFF
--- a/include/boost/config/detail/suffix.hpp
+++ b/include/boost/config/detail/suffix.hpp
@@ -528,12 +528,6 @@ namespace boost {
 #  define BOOST_APPEND_EXPLICIT_TEMPLATE_NON_TYPE(t, v)
 #  define BOOST_APPEND_EXPLICIT_TEMPLATE_NON_TYPE_SPEC(t, v)
 
-// When BOOST_NO_STD_TYPEINFO is defined, we can just import
-// the global definition into std namespace:
-#if defined(BOOST_NO_STD_TYPEINFO) && defined(__cplusplus)
-#include <typeinfo>
-namespace std{ using ::type_info; }
-#endif
 
 // ---------------------------------------------------------------------------//
 
--- a/include/boost/config/stdlib/dinkumware.hpp
+++ b/include/boost/config/stdlib/dinkumware.hpp
@@ -86,16 +86,6 @@
 #  define BOOST_NO_STD_LOCALE
 #endif
 
-// Fix for VC++ 8.0 on up ( I do not have a previous version to test )
-// or clang-cl. If exceptions are off you must manually include the 
-// <exception> header before including the <typeinfo> header. Admittedly 
-// trying to use Boost libraries or the standard C++ libraries without 
-// exception support is not suggested but currently clang-cl ( v 3.4 ) 
-// does not support exceptions and must be compiled with exceptions off.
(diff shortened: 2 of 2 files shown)


include/boost/config/detail/suffix.hpp
```diff
@@ -528,12 +528,6 @@ namespace boost {
 #  define BOOST_APPEND_EXPLICIT_TEMPLATE_NON_TYPE(t, v)
 #  define BOOST_APPEND_EXPLICIT_TEMPLATE_NON_TYPE_SPEC(t, v)
 
-// When BOOST_NO_STD_TYPEINFO is defined, we can just import
-// the global definition into std namespace:
-#if defined(BOOST_NO_STD_TYPEINFO) && defined(__cplusplus)
-#include <typeinfo>
-namespace std{ using ::type_info; }
-#endif
```
jeremy-murphy, line 536: I have to manually delete this block of code to get Boost.Config to compile with the IAR compiler, so I hope this PR gets merged soon!

jeremy-murphy: As I mentioned on the issue tracker, I'm not sure removing these lines is the best solution after all.

jzmaddock: This was an MSVC-specific workaround as I recall, if I restrict the code to MSVC does that help?

BTW is there really no typeinfo?  What else is missing?

jeremy-murphy: Really and truly. Granted, this is _because_ I have disabled exceptions and RTTI on this compiler because our target is embedded. So for example I have `BOOST_NO_EXCEPTIONS`, `BOOST_NO_RTTI` and `BOOST_NO_TYPEID` defined in the preprocessor, and that solves most anticipated problems.

jeremy-murphy:
@Kojoley I think you need to resolve the merge conflicts for this to ever have a hope of being merged.

Kojoley:
The change it conflicts with (035c096da942a3fa7f75180ae8b2f2732db0a9f4) is just wrong and should be reverted, then this will land cleanly.

jzmaddock:
The problem I have is that this PR breaks the original purpose of those workarounds.

Try this: https://github.com/boostorg/config/commit/46e0a134966bebb5496a2fd8d26a101e4b94ca95

Kojoley:
What is `the original purpose of those workarounds`? To slow compilation for everyone because of Boost.LexicalCast? I fixed Boost.LexicalCast and it is not need the workaround.

PR #308, opened 2019-12-03
Use <version> to determine standard library

Kojoley:


jzmaddock:
I'm not necessarily against this (as long as it works OK), but note that <cstddef> is a pretty tiny header already, so I doubt this change will have much impact?

Kojoley:
Compile time impact is very tiny (clang+msvc stdlib is 0.012 vs 0.007), and IIUC `<version>` should be included with any other standard library header. On the other hand, including less things from standard library will force self-sustainability of headers that include `boost/config.hpp` but forgets to include `cstddef`.

PR #310, opened 2019-12-12
Added new defect macro for C++11's unrestricted unions.

eldiener:


eldiener:
I would like merge this PR unless there are any objections.

PR #314, opened 2019-12-24
Fix Boost library list links to Config documentation

glenfe:


eldiener:
Shouldn't this just be "index.html" rather than "doc/html/index.html" ?

glenfe:
Going directly to "doc/html/index.html" saves a redirect. Ideally more of the meta/libraries.json would do this, such that the generated libraries.htm pages have fewer redirects. 

e.g. https://www.boost.org/doc/libs/master/libs/libraries.htm#IO    The link to Boost.IO avoids the redirecting index.html and goes directly to IO's documentation.

eldiener:
All other libraries in a Boost distribution have the link to their docs in the top level directory via an index.html. There is no reason config should be different.

glenfe:
Config is no different; it has an index.html which has that same kind of redirect. But there are Boost libraries which specify "documentation" in meta/libraries.json so that all relevant links can bypass that index.html redirect. i.e. Not *all* Boost Libraries avoid the documentation: element in meta/libraries.json

glenfe:
Here are some of the Boost libraries that specify a documentation tag in meta/libraries.json directly to the documentation (avoiding redirect HTML files):
https://github.com/search?utf8=%E2%9C%93&q=org%3Aboostorg+filename%3Alibraries.json+documentation&type=Code

- Tribool
- Interval
- IO
- Fusion
- DynamicBitset
- Exception

To list a few. (i.e. Not quite "All other").

PR #316, opened 2020-01-01
Add BOOST_INLINE_CONSTEXPR convenience macro

glenfe:


glenfe:
@jzmaddock , I'm guessing you're busy. Any objections to this being merged?

jzmaddock:
None but...

* How useful is this if inline variables are not supported - do we not end up with multiple definitions?
* constexpr implies inline, so isn't declaring a variable simply as BOOST_CONSTEXPR_OR_CONST the same thing anyway?

glenfe:
At file scope constexpr does not imply inline linkage, it is internal linkage. 

So, for example: https://github.com/boostorg/core/blob/develop/include/boost/core/empty_value.hpp#L142-L148 

This would have internal linkage and have a definition in each translation unit. Which is fine because it's the best we can do pre-C++17.

This is actually the use case which prompted Andrey to add BOOST_INLINE_VARIABLE in the first place.

jzmaddock:
OK, best merge this then....

PR #328, opened 2020-04-05
Further changes for Embarcadero C++ clang-based compilers

eldiener:
I discovered through testing some issues and made further changes to the Embarcadero C++ clang-based compilers support in config. I also updated the test for support of the cwchars header file involving a basic test for multi-byte character support in the header file. The Embarcadero C++ clang-based compilers were passing the test but were actually missing all the multi-byte support in the header file, including all wide character to multi-byte and all the multi-byte to wide character functionality in the header file. So I felt updating the test slightly just to make sure that multi-byte support existed in its most basic form was necessary. I do not think this will affect any other compilers, and in my testing all the compilers I tested were not affected in any way by failing the test where previously they may have passed it.

### DIFF
--- a/include/boost/config/compiler/codegear.hpp
+++ b/include/boost/config/compiler/codegear.hpp
@@ -111,30 +111,6 @@
 #  endif
 #endif /* defined(BOOST_EMBTC) */
 
-#if defined(BOOST_EMBTC_WINDOWS)
-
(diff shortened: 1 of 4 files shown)


eldiener:
Any objections to merging this PR ?

jzmaddock:
Nope, merging now....

PR #331, opened 2020-04-17
Atomic exchange for 16 bytes does not work for Embarcadero C++ clang-…

eldiener:
…based compilers.


eldiener:
Any objections merging this to 'develop' ?

PR #333, opened 2020-05-01
Changes for Embarcadero C++ clang-based compilers, targeting Boost 1.74. Documented the basic Embarcadero macros. I did not document the exten…

eldiener:
…ded Embarcadero macros beginning with BOOST_EMBTC_ but I did define them all to the value of 1 whereas before they were just defined to an empty value.


eldiener:
Are there any objections to merging this PR ?

PR #338, opened 2020-07-24
Target attribute definition

nemothenoone:
Target attribute definition introduces function multi-versioning.

### DIFF
--- a/include/boost/config/compiler/clang.hpp
+++ b/include/boost/config/compiler/clang.hpp
@@ -351,6 +351,6 @@
 // Macro used to identify the Clang compiler.
 #define BOOST_CLANG 1
 
-#if (__clang_major__ >= 3) && (__clang_minor__ >= 8)
+#if (__clang_major__ >= 4 || (__clang_major__ >= 3 && __clang_minor__ >= 8))
 #  define BOOST_ATTRIBUTE_TARGET(isa) __attribute__ ((target(isa)))
 #endif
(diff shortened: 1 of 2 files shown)


jzmaddock:
What's the motivation for adding this to Boost.Config rather than the specific library that needs it - or is there a need across multiple libraries?

nemothenoone:
Hm. Good question. I was thinking of some kind of architecture-specific function selection macro for Boost.Multiprecision, Boost.Math and for one more library coming. I think lots of architecture-specific places (e.g. with intrinsics) across the whole Boost suite could use this macro.

jzmaddock:
My inclination is to not merge this at this time - the rational is:

* It's completely compiler specific (not necessarily a show stopper).
* It's not required by multiple Boost libraries yet.

But... I can see the logic so let's leave it open for now.

What tends to happen with rather specific stuff like this, is they get trialled in a particular library first, and then moved here when more than one library needs them, or else when the configuration is clearly more easily done via the logic that Boost.Config already has.

With regard to multiprecision, what we've done so far is just tip a toe into the water of using intrinsics, but these are abstracted so that the actual code doesn't have to deal with 32/64 bit issues, or compiler specific stuff (ie works with msvc as well as gcc): https://github.com/boostorg/multiprecision/blob/develop/include/boost/multiprecision/cpp_int/intel_intrinsics.hpp.

PR #341, opened 2020-08-14
Bump version to 1.75.0

mclow:


jzmaddock:
Please go ahead and merge when the checks are done (not that I can imagine anything that would fail!).

PR #346, opened 2020-11-25
Don't add address model in Embarcadero toolset name

tanzislam:
None of the other toolset names include a 32/64-bit indicator. Remove the `32`/`64` and change the name to `embtc` to match with boostorg/build#675, so that auto-linking actually works.


jzmaddock:
@eldiener ?

eldiener:
I am fine with this change.

PR #372, opened 2021-03-12
Add feature test macros for C++11 features.

Breakthrough:
This PR adds some feature test macros for the C++11 features.  I wasn't sure about decltype, so left that as-is for now, but documented what the appropriate check should be.

### DIFF
--- a/include/boost/config/compiler/common_edg.hpp
+++ b/include/boost/config/compiler/common_edg.hpp
@@ -77,31 +77,20 @@
 
 #define BOOST_NO_CXX11_AUTO_DECLARATIONS
 #define BOOST_NO_CXX11_AUTO_MULTIDECLARATIONS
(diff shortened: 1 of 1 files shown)


jzmaddock:
That looks fair enough.... merging.

Yes your comment about __decltype looks correct to me, __cpp_rvalue_references for BOOST_NO_CXX11_RVALUE_REFERENCES, there may be others.

Are you testing on a specific compiler?  I didn't think many were using the EDG frontend these days.

Breakthrough:
Specifically GreenHills in this case.  I initially considered just modifying that compiler header, but then I noticed that the C++14 feature test macros were already accounted for in the EDG-common checks, so figured might as well update all the C++11 ones to match.  This allows a few more boost libraries to compile out of the box now.

Thanks so much for the quick response, as well as for your work on boost!!

PR #375, opened 2021-03-16
Fix BOOST_NO_CXX20_HDR_RANGES

mborland:
Corrects minor typo in the name of the header.


mborland:
@jzmaddock This should be good to go. Failures look similar to https://github.com/boostorg/math/issues/572

PR #378, opened 2021-04-01
Tentative fix for CUDA/Clang/Cray issue reported here https://github.…

jzmaddock:
…com/boostorg/math/issues/181.


msimberg:
Works at least in our case at least and looks reasonable to me. Thank you!

PR #386, opened 2021-07-04
Minor fix for gcc-4.4 with no rtti.

jzmaddock:


jzmaddock:
On second thoughts, this is unfixable - apparently you cannot include <memory> in C++0x mode with -fno-rtti.

pdimov:
It's fixable in the .drone.star file by removing 0x from g++ 4.4. :-)

jzmaddock:
> It's fixable in the .drone.star file by removing 0x from g++ 4.4. :-)

Yes, I know, but it's really only the one test that's effected.  When I get a chance I'll disable that one test.

PR #254, opened 2018-12-02
Unify mips archtectures into mips

FlyGoat:

PR #389, opened 2021-07-22
Rename mips1 to mips

wzssyqa:
In fact mips1 is now used for all mips revisions.


wzssyqa:
this PR is duplicated with https://github.com/boostorg/config/pull/254.
So close it.

Issue #392, opened 2021-08-06
Conflict between BOOST_NOINLINE and HIP compiler


markdewing:
The hipcc compiler (tested on rocm 4.3) defines `__noinline__` as `__atribute__((noinline))` (in /opt/rocm-4.3.0/hip/include/hip/amd_detail/host_defines.h)
This definition mixes with the boost definition to create an incorrect `__attribute__` statement
Shows up when including stacktrace.hpp

Reproducer
```
// Workaround
//#define BOOST_NOINLINE __attribute__((noinline))__

#define BOOST_STACKTRACE_USE_BACKTRACE
#include <stacktrace.hpp>

#include <hip/hip_runtime.h>

int main()
{
    return 0;
}

BOOST_NOINLINE void stuff()
{
    int i = 1;
}
```

Compile with hipcc and get the following output
```
boost_test.cpp:13:1: error: use of undeclared identifier 'noinline'; did you mean 'inline'?
BOOST_NOINLINE void stuff()
^
/usr/include/boost/config/detail/suffix.hpp:623:46: note: expanded from macro 'BOOST_NOINLINE'
#      define BOOST_NOINLINE __attribute__ ((__noinline__))
                                             ^
/opt/rocm-4.3.0/hip/include/hip/amd_detail/host_defines.h:50:37: note: expanded from macro '__noinline__'
#define __noinline__ __attribute__((noinline))
                                    ^
boost_test.cpp:13:1: error: type name does not allow function specifier to be specified
/usr/include/boost/config/detail/suffix.hpp:623:46: note: expanded from macro 'BOOST_NOINLINE'
#      define BOOST_NOINLINE __attribute__ ((__noinline__))
                                             ^
/opt/rocm-4.3.0/hip/include/hip/amd_detail/host_defines.h:50:37: note: expanded from macro '__noinline__'
#define __noinline__ __attribute__((noinline))
                                    ^
boost_test.cpp:13:1: error: expected expression
/usr/include/boost/config/detail/suffix.hpp:623:46: note: expanded from macro 'BOOST_NOINLINE'
#      define BOOST_NOINLINE __attribute__ ((__noinline__))
                                             ^
/opt/rocm-4.3.0/hip/include/hip/amd_detail/host_defines.h:50:46: note: expanded from macro '__noinline__'
#define __noinline__ __attribute__((noinline))
                                             ^
3 errors generated when compiling for gfx900.
```

[...]

PR #393, opened 2021-08-08
Patch for noinline attribute on the HIP compiler.

jzmaddock:
Fixes https://github.com/boostorg/config/issues/392.


atamazov (commented):
I recommend re-implementing the fix.

PR #394, opened 2021-08-08
Add std::is_constant_evaluated support.

jzmaddock:


doc/macro_reference.qbk
```diff
@@ -1081,6 +1082,33 @@ that are not yet supported by a particular compiler or library.
 
 [endsect]
 
+[section Macros that allow C++20 features to be used in previous standard versions]
+
+[table
+[[macro][Description]]
+[[`BOOST_IS_CONSTANT_EVALUATED()`][Evaluated to `true` when the current context is `constexpr` and false otherwise.
```
libbooze, line 1089: beginner question: why is this a macro?

I think boost::is_constant_evaluated() would be a bit nicer, I presume this is for consistency with other stuff in this repo?

doc/macro_reference.qbk
```diff
@@ -1081,6 +1082,33 @@ that are not yet supported by a particular compiler or library.
...
+                                 The macro may be functional in C++14 onwards if the compiler has a suitable intrinsic, in particular this macro is functional for gcc-6 and onwards
+                                 as well as for compilers which have direct support for `std::is_constant_evaluated`.
+
+                                 Note that before using this macro, the program must include `<type_traits>`.
+
+                                 Note that this macro may behave subtly different from `std::is_constant_evaluated()` in that it may evaluate to
+                                 `true` when code is evaluated in a non-constexpr context, but is none the less evaluated at "compile time" as a result
+                                 of program optimisation.
+
+                                 The macro always evaluates to `false` on unsupported/old compilers.]]
```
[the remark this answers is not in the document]
jzmaddock: Nod.  There will certainly be tricky cases where you can abuse it into doing the wrong thing, if it's easier I could just omit that macro.

doc/macro_reference.qbk
jzmaddock: We would need to include <type_traits> to cover that, and I'd very much like to NOT do that.

Explanation: if it's a compiler about which we know nothing about, ie we don't know what it calls the compiler intrinsic that implements this, then we would need to be able to forward to std::is_constant_evaluated if it's presence is advertised in <version>.

As it stands, it's up to users of these macros to include <type_traits> and everyone else can do without if they wish and only pay for what they use.

We could of course put a real function into Boost.Core.

doc/macro_reference.qbk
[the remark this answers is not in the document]
libbooze: >  if it's easier I could just omit that macro.

Hard for me to judge.
My suggestion is to ask the users what they feel. I would personally never use it since I consider it too tricky, but I am not really an expert library developer and I have no usecases for it.

doc/macro_reference.qbk
libbooze: Ah, I see, I was naively assuming that that feature macro is predefined, now I see it is in `<type_traits>`. Shame.

test/test_boost_is_constant_evaluated.cpp
```diff
@@ -0,0 +1,60 @@
+//  (C) Copyright John Maddock 2021.
```
libbooze, line 1: Would it be good to add a "sanity test"?

For some compiler you know it should have proper machinery in C++20 mode? I am not sure if boost tests have modern compilers running with -std=c++20, but I was thinking about something like this.

https://www.godbolt.org/z/r9Y5x3x6M

(use of assert just as an example, it can be a test framework assert_eq in real code)

doc/macro_reference.qbk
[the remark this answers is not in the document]
Lastique: `__builtin_constant_p(n)` is not a constant expression, and it does not indicate whether `n` is a constant expression either. See [here](https://gcc.godbolt.org/z/3h1KP9oK8). What it shows is whether the compiler is able to deduce the value of `n` at compile time, including as a result of optimizations (e.g. inlining). This means that its value may change depending on optimization level used. If it returns `true`, `n` may not be a constant expression in C++ definition and thus may not be used in C++ constant expressions.

I think `BOOST_IS_CONSTANT_EVALUATED_INT` needs to be renamed to avoid confusion with `std::is_constant_evaluated`. `__builtin_constant_p` is useful, but it is a very different thing.

PR #414, opened 2022-01-20
Add BOOST_NO_CXX17_DEDUCTION_GUIDES config macro

Lastique:
The macro allows to detect whether the compiler supports C++17 class template argument deduction (CTAD) guides.

This macro is needed at least in Boost.Atomic.


jzmaddock:
Sorry for being late coming to this, it's a good idea, but since the logic is the same everywhere, should we not just add the logic once to suffix.hpp and leave the individual compiler configs unchanged (if there are compiler bugs which need workarounds we can always over ride at the individual compiler level).

jzmaddock:
Superseded by https://github.com/boostorg/config/pull/433

PR #421, opened 2022-02-03
Refactor C++20 header configuration.

jzmaddock:


pdimov:
You can fix the `mt.exe` errors by adding `embed-manifest-via=linker` to the `b2` command line. `clang-win.jam` doesn't detect the path to `mt.exe` properly for the Windows 10 Platform SDK, and even though I wrote `clang-win.jam`, I don't quite know how to fix it, but embedding via the linker works. :-)

include/boost/config/detail/suffix.hpp
```diff
@@ -1198,6 +1198,14 @@ namespace std{ using ::type_info; }
 #endif
 #endif
 
+#ifdef __has_include
+#if !__has_include(<version>)
+#  define BOOST_NO_CXX20_HDR_VERSION
+#endif
```
pdimov, line 1208: I think that it would be useful to have an `#else` here that does `#include <version>` so that we can rely on it being always included if available, even if `select_stdlib_config` didn't include it because of a user override.

PR #431, opened 2022-05-22
Add BOOST_NULLPTR helper macro

pdimov:
Compilers and static analysis tools have started warning on the use of integral null pointer constants, suggesting that they need to be replaced with `nullptr`. This leads to people submitting pull requests that replace the uses of `0` with `nullptr`, such as for example https://github.com/boostorg/throw_exception/pull/22 (https://github.com/boostorg/throw_exception/pull/22/commits/8678b7cb09ad71bffbc24b099fc2a86bd925046a). However, `nullptr` is not available in C++03, which necessitates checking `BOOST_NO_CXX11_NULLPTR` as in https://github.com/boostorg/throw_exception/pull/22/commits/0a752cc9f87be1faaf8d7adcbc2cf4220623fbc1.

The proposed helper macro `BOOST_NULLPTR` expands to `nullptr` when it's supported, `0` otherwise, which avoids the need to check `BOOST_NO_CXX11_NULLPTR` in libraries themselves.

It would be more correct to expand this to `NULL` when `nullptr` isn't available, as `NULL` has special meaning for GCC and doesn't cause the warning, even in C++03 mode. But this requires that `<stddef.h>` be included in `suffix.hpp`, and I wasn't sure that this was acceptable.

### DIFF
--- a/doc/macro_reference.qbk
+++ b/doc/macro_reference.qbk
@@ -966,6 +966,10 @@ and must be defined for all translation units in the program, including Boost li
 This macro will no longer have any effect once an official Microsoft
 release supports the CTP features.
 ]]
+[[`BOOST_NULLPTR`][
+If `BOOST_NO_CXX11_NULLPTR` is not defined (i.e. C++11 compliant compilers), expands to `nullptr`,
+otherwise expands to `0`.
+]]
 ]
 
 [endsect]
--- a/include/boost/config/detail/suffix.hpp
+++ b/include/boost/config/detail/suffix.hpp
@@ -1069,6 +1069,12 @@ namespace std{ using ::type_info; }
 
 #define BOOST_STATIC_CONSTEXPR  static BOOST_CONSTEXPR_OR_CONST
 
+#if !defined(BOOST_NO_CXX11_NULLPTR)
+# define BOOST_NULLPTR nullptr
+#else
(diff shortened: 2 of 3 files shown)


pdimov:
Well?

glenfe:
Looks good to me, let's see what John thinks.

jzmaddock:
Sorry to be late to this... busy with tax returns :(

I'm fine with <cstddef> being included if it's done only when needed ie in C++03.

The failing drone CI runs look like their no-longer supported configurations.

glenfe:
Yes, only those Boost libraries that support C++03 would use BOOST_NULLPTR, the rest would just use nullptr.

glenfe (approved):

Flamefire:
> It would be more correct to expand this to NULL when nullptr isn't available, as NULL has special meaning for GCC and doesn't cause the warning, even in C++03 mode. But this requires that <stddef.h> be included in suffix.hpp, and I wasn't sure that this was acceptable.

I would have done that and conditionally include it (for C++03). IMO it is better to be (more) correct here and that header is likely one of the lightest and most likely will be included anyway at least transitively.

pdimov:
Would be nice to have this merged to master as well.

jzmaddock:
Done for both.

PR #441, opened 2022-08-02
Add a macro BOOST_DISABLE_EXPLICIT_SYMBOL_VISIBILITY to disable mandatory symbol visibility

jcelerier:


nigels-com:
We'd use that, for sure.  +1

mardigontoler:
This would be useful for me, as well.

jcelerier:
any possible review ? @pdimov @jzmaddock

pdimov:
Should probably be done in suffix.hpp once; if defined, undefine the other macros and define them as empty. Here for instance: https://github.com/boostorg/config/blob/1cff5e37bb69f42ffab73438de61e34c71258f19/include/boost/config/detail/suffix.hpp#L40

jcelerier:
done

jzmaddock:
Looks good to me, merging...

jzmaddock:
Ah, actually I see this breaks the documentation build, looks like there's a ']' missing.

Kojoley:
Wouldn't linker strip out all the symbols and leave you with an empty static library?

jcelerier:
@Kojoley no, why would it ? static libraries contain all the (non-static/anonymous namespace) symbols.

Kojoley:
> @Kojoley no, why would it ? static libraries contain all the (non-static/anonymous namespace) symbols. actually it doesn't really make sense as a question: the linker isn't involved in making static libraries,

https://gcc.gnu.org/onlinedocs/gnat_ugn/Compilation-options.html

> they are just an archive of the .o's

No, archives and static libraries are different things https://softwareengineering.stackexchange.com/questions/389383/what-is-the-difference-between-a-static-library-and-an-archive-library

jcelerier:
.. the stackoverflow link you pasted literally says "A static library is just an ar archive (like zip but ancient) of object files."

Issue #280, opened 2019-04-28
add BOOST_FUNCTION_DEPRECATED


HDembinski:
It would be great to have a cross-platform way give a deprecation warning at compile-time for pre-c++14 compilers. In C++14, there is `[[deprecated]]`, but older compilers already supported this with compiler-specific extensions, see e.g.:

https://stackoverflow.com/questions/295120/c-mark-as-deprecated/21265197#21265197

If it is hard to do, maybe not worth investing the time, but it could be very useful.

Lastique:
See https://github.com/boostorg/config/pull/442.

PR #442, opened 2022-08-14
Add a BOOST_DEPRECATED macro for deprecated symbol markup

Lastique:
`BOOST_DEPRECATED` can be used to mark functions, types and objects as deprecated, with a message with a recommendation of replacement. Using such marked symbols in code will generate compiler warnings, with the
specified message, if possible.

The warnings can be suppressed if `BOOST_ALLOW_DEPRECATED_SYMBOLS` is defined. Additionally, added support `BOOST_ALLOW_DEPRECATED` macro that not only allows for deprecated symbols but also deprecated headers (i.e. defining
`BOOST_ALLOW_DEPRECATED` is equivalent to defining both `BOOST_ALLOW_DEPRECATED_SYMBOLS` and `BOOST_ALLOW_DEPRECATED_HEADERS`).

Closes https://github.com/boostorg/config/issues/280.

### DIFF
--- a/doc/macro_reference.qbk
+++ b/doc/macro_reference.qbk
@@ -1199,41 +1199,41 @@ struct foo{
 Normally evaluates to nothing, but evaluates to return x; if the compiler
 requires a return, even when it can never be reached.
 ]]
-[[`BOOST_FALLTHROUGH`][ 
-The BOOST_FALLTHROUGH macro can be used to annotate implicit fall-through 
-between switch labels: 
-`` 
- switch (x) { 
- case 40: 
- case 41: 
-    if (truth_is_out_there) { 
-       ++x; 
-       BOOST_FALLTHROUGH;  // Use instead of/along with annotations in 
-       // comments. 
-    } else { 
-      return x; 
-    } 
-    case 42: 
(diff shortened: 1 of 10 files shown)


Flamefire (approved):
Yes please, this is long overdue!

Looks good to me.

Lastique:
Gentle ping @jzmaddock.

Also CC @pdimov as this touches `BOOST_HEADER_DEPRECATED`.

pdimov:
I'm OK with that addition to `BOOST_HEADER_DEPRECATED`.

Lastique:
I would prefer to leave the message up to the users of the macro. There may be use cases where the mechanism of disabling the warnings is different from `BOOST_ALLOW_DEPRECATED(_SYMBOLS)`. For example, if the library has its own local config macro for disabling warning only from that particular library.

Besides, modifying the message implies a specific format or wording of the message specified by user (i.e. does it start with a capital? does it end with a dot so that the instructions start with a new sentence? does it end with a space? etc.) I don't want to make such assumptions.

jzmaddock:
OK, lets go with this.

PR #444, opened 2022-08-29
Update Clang version detection

pdimov:
1. Use `__clang_patchlevel__ % 100`, because `__clang_patchlevel__` is sometimes higher than 100 and the version comes out incorrect. (Reported in https://github.com/boostorg/assert/issues/28.)
2. Update Apple Clang version detection for XCode 13.3, 13.4, 14.0 from https://en.wikipedia.org/wiki/Xcode#Toolchain_versions.


jzmaddock:
Looks good to me, one failure so far is a spurious clone-failure, please go ahead and merge when you're happy!

PR #452, opened 2022-11-09
Add macro BOOST_MSSTL_VERSION

pdimov:
... as [suggested in 2018](https://lists.boost.org/Archives/boost/2018/07/242452.php). (Better late than never I guess.)


jzmaddock:
@pdimov can you add some docs as well please?

pdimov:
Added documentation.

While doing so noticed that the same table contains `BOOST_NO_WREGEX`, documented to be defined in `boost/regex.hpp`. Is this some sort of a leftover? It seems out of place here.

jzmaddock:
>While doing so noticed that the same table contains BOOST_NO_WREGEX, documented to be defined in boost/regex.hpp. Is this some sort of a leftover? It seems out of place here.

Definitely a leftover, will remove.

PR #456, opened 2022-11-15
Declare BOOST_ATTRIBUTE_UNUSED to be [[maybe_unused]] in C++17

mojca:
See https://github.com/boostorg/serialization/issues/271 and https://github.com/boostorg/serialization/issues/145

I'm getting a compiler warning (or rather error since I try to compile with "treat warnings as errors") in Visual Studio 2022.

Since this became a standard in C++17 (https://en.cppreference.com/w/cpp/language/attributes/maybe_unused), it might be worth using that one. The compiler warning disappears.


jzmaddock:
I fixed this slightly differently in the referenced PR.

PR #457, opened 2022-11-16
Expand BOOST_ATTRIBUTE_UNUSED [[maybe_unused]]

jzmaddock:
When available, supersedes: https://github.com/boostorg/config/pull/456


mojca:
Thank you very much

mojca:
@jzmaddock: is there any reason why this change didn't make it into boost 1.81?

jzmaddock:
For reasons I don't understand it's not in the 1.81 zip.  Possibly I was late in merging to master, not sure about that, and Git isn't telling me the answer.

Issue #462, opened 2022-12-29
Missing define BOOST_NO_CXX11_NULLPTR on OSX with libstdc++


Flamefire:
A build on XCode 9.4.1 on [Drone](https://drone.cpp.al/boostorg/boost-ci/341/38/2) fails because the standard library (stdlibc++) is lacking the `std::nullptr_t` type.

I believe this means that `BOOST_NO_CXX11_NULLPTR` should be defined but is not. It seems to be possible that the compiler supports the `nullptr` keyword but just that the standard library is to old.

See also https://github.com/CPPAlliance/drone-ci/issues/15

jzmaddock:
Now fixed in develop.  I did not fix up C++17 compiler support though - the ancient std lib will likely cause unfixable issues.

Flamefire:
Thanks!

PR #463, opened 2022-12-29
Add XCode 9.4.1 testing.

jzmaddock:
Refs https://github.com/boostorg/config/issues/462.


Flamefire:
Wouldn't the C++17 test be good way to ensure nothing is missed? E.g. one test fails due to missing `<tuple>` hence either there is a missing check of `BOOST_NO_CXX11_HDR_TUPLE` or it is wrongly not defined.

I mean the purpose of Boost.Config (those macros) are to detect defects of the kind "not supported by the std lib" so IMO that test is valuable.

PR #473, opened 2023-04-02
Fix assert_cxx03.hpp and assert_cxx98.hpp

Kojoley:


Kojoley:
I don't want to think how this should be resolved, feel free to just close the PR.

jzmaddock:
This doesn't need fixing IMO: those C++98 macros get set for later (ie C++14) versions which have deprecated and removed those features.  What I'll do is remove the undocumented assert_cxx98.hpp file which should never have been there.

See https://github.com/boostorg/config/pull/546.

PR #487, opened 2023-10-20
Use `__apple_build_version__` to detect Apple clang

chrstphrchvz:
…since `__APPLE__` is defined by both Apple Clang and LLVM.org Clang on Apple platforms (as mentioned by https://github.com/boostorg/config/pull/39#issuecomment-59946538).

This ensures `BOOST_CLANG_VERSION` is defined correctly when using LLVM.org clang.
| Compiler               | Before | After  |
|------------------------|--------|--------|
| Xcode 14.2 Apple Clang | `140000` | `140000` |
| LLVM.org Clang 17.0.3  | `140000` | `170003` |
| LLVM.org Clang 9.0.1   | `40000`  | `90001`  |

### DIFF
--- a/include/boost/config/compiler/clang_version.hpp
+++ b/include/boost/config/compiler/clang_version.hpp
@@ -2,7 +2,7 @@
 // Distributed under the Boost Software License, Version 1.0.
 // https://www.boost.org/LICENSE_1_0.txt)
 
-#if !defined(__APPLE__)
+#if !defined(__apple_build_version__)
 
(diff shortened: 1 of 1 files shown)


pdimov:
While we're here we should probably update the version check for the newer Xcode releases, as in https://github.com/boostorg/config/pull/488.

They had the perfect opportunity to synchronize their version numbers with upstream LLVM starting with 14.0 and they of course didn't take it.

(Unrelated, we might want to add macos-13 to GHA.)

PR #491, opened 2023-12-14
Update version.hpp to 1.85.0

mclow:


pdimov:
May we please have this merged? To master as well. Jamroot and Config versions being out of sync causes issues.

PR #492, opened 2024-01-22
Fix Boost.Filesystem usage, add `BOOST_NO_CXX17_AUTO_NONTYPE_TEMPLATE_PARAMS`

Lastique:
This PR:

* Updates `tools/generate.cpp` to avoid using Boost.Filesystem APIs that were deprecated and now removed.
* Adds a new `BOOST_NO_CXX17_AUTO_NONTYPE_TEMPLATE_PARAMS` macro that indicates whether the compiler does not support C++17 [`auto` non-type template parameters](https://en.cppreference.com/w/cpp/language/template_parameters).


Lastique:
Ping @jzmaddock. Note that at least the first commit needs to be merged, as currently `tools/generate.cpp` fails to compile.

jzmaddock:
Failure are due to b2 dropping GCC-4 support, merging...

Lastique:
Thanks. Drone failure is because of a `b2` [issue](https://github.com/bfgroup/b2/issues/360), which was reportedly fixed.

Lastique:
Please, also merge to master as there is downstream code that uses this macro that I would like to merge as well.

jzmaddock:
Done.

PR #504, opened 2024-11-10
Add `BOOST_FALLTHROUGH` for MSVC

Flamefire:
MSVC supports this attribute for C++17 and higher since VS 2017: https://learn.microsoft.com/en-us/cpp/cpp/attributes?view=msvc-170#fallthrough

Not sure if it is worth adding a check for `_MSC_VER >= 1910` as it isn't done for `BOOST_ATTRIBUTE_UNUSED` either


Flamefire:
@glenfe Can we include this in 1.89?

PR #510, opened 2025-02-14
Fixes Autolinking on new boost version

### LINKED ISSUE #496: VS 2022 v17.10
VS2022 v17.10 will be officially released soon. _MSC_VER is 1940 so it will not be mapped to 14.3 in boost code.

https://github.com/boostorg/config/blob/5734e160e08b8df898c7f747000f27a3aafb7b2b/include/boost/config/compiler/visualc.hpp#L375C1-L377C9

https://devblogs.microsoft.com/cppblog/msvc-toolset-minor-version-number-14-40-in-vs-2022-v17-10/

OlekRaymond:
Fixes #496


neusdan:
I think that is not the correct fix because MSVC 19.40 is still `vc143`. I think #497 is more appropriate.

jzmaddock:
I'm looking into this now, we need to do *exactly* the same thing as Boost.Build.

jzmaddock:
So far as I can tell from current develop, I still get "143" in library names, I haven't been able to spot where Boost.Build sets this though.  Attempting to build with toolset=msvc-14.4 produces:

```
error: "14.4" is not a known subfeature value of <toolset>msvc
```

Can you confirm?

neusdan:
> , I still get "143" in library names

That's weird because if i set `toolset=msvc-14.4` then it will produce library names with `vc144` in it. 

Not sure if it is related but i am also calling `vsdevcmd.bat -vcvars_ver=14.4` before calling `b2.exe`.

OlekRaymond:
> I still get "143" in library names
This is what happened to us.

I have experienced oddities with setting `vcvars`, we just used the default for the latest version then downgraded to get around this issue

jzmaddock:
>That's weird because if i set toolset=msvc-14.4 then it will produce library names with vc144 in it.

I don't see how you can do that, I checked the source (https://github.com/boostorg/build/blob/develop/src/tools/msvc.jam) and there is no such thing as a "14.4" version, and for me attempting to use it, generates a hard error from b2 (which it should).  This looks like user error to me since we only support 14.3 as a valid version not 14.4.  

I don't know how you managed to slip by the error checks though.

@grafikrobot would you have any idea what's happening at the b2 end of things?

jzmaddock:
So I just updated to the latest and greatest VS-17.14, which has msvc version 194435217.  For me Boost.Build is still identifying that as msvc-14.3 and naming binaries appropriately.  However.... the update has mangled something badly, as the IDE will no longer build C++ programs (that's the second update in a row that's done that, I'll have to completely remove and reinstall to fix - when I have the time).

Meanwhile I'd still like to hear from @grafikrobot what's intended in Boost.Build, as ultimately the only thing that matters is that we do the same thing.

PR #517, opened 2025-05-06
Tentatively fix Circle-ci build.

jzmaddock:


jzmaddock:
Remaining failure is a network issue on drone, merging.

.circleci/config.yml
```diff
@@ -6,7 +6,7 @@ jobs:
       - BOOST_LIBRARY=config
       - CXX_STANDARD=gnu++11
     docker:
-      - image: gcc:7
+      - image: gcc:13
```
grafikrobot, line 9: Where does one find what images are available? I spent many minutes reading CircleCI docs and I didn't find gcc or clang specific images.

.circleci/config.yml
```diff
@@ -46,7 +46,7 @@ jobs:
       - run:
           name: Building inspect
           command: |
-            cd $BOOST/boost/tools/inspect/build && ../../../b2 -j2 address-model=64 architecture=x86 toolset=gcc cxxflags="-std=gnu++14" release dist-bin
+            cd $BOOST/boost/tools/inspect && ../../b2 -j2 address-model=64 architecture=x86 toolset=gcc cxxflags="-std=gnu++14" release dist-bin
```
grafikrobot, line 49: It would be better to use the `cxxstd=14` feature instead of the flag directly.

.circleci/config.yml
jzmaddock: I believe they are the official docker images rather than circleci ones.  But I genuinely have no idea.

.circleci/config.yml
jzmaddock: Good point, fossilized old code lurking, I'll update next time I'm in there.

PR #522, opened 2025-09-19
Fixes CI errors in GHA, Appveyor and Drone

anarthal:
* Updates actions/checkout from v3 to v5
* Removes the ci-skip action, as it's now built into GHA
* Removes the Windows 2019 builds from GHA, since these runners are no longer available
* Adds a MSVC 14.2 build to Appveyor, replacing those removed from GHA
* Adds missing dependencies to the Appveyor build
* Removes the --experimental-wasm-threads flag from the wasm builds, which was removed in Node 20.0
* Updates the Drone gcc-13 build to use Ubuntu 24.04, which is LTS


anarthal:
I do agree, since Appveyor is painfully slow. But the migration is not trivial for me, as I've never run MSVC builds before 14.0 in Drone before.

The only build which was failing now was a MSVC 14.1 that I added to Appveyor for completeness, which we didn't use to have. I've removed it, returning to the status quo, and this should (hopefully) work.

My proposal is merging this once everything is green, so new changes like the updated Boost version header can be merged, and I'll submit another PR to get rid of Appveyor once I discover the best way to do it.

PR #525, opened 2025-11-02
Add GetSystemTimePreciseAsFilename

pgawro:
According to Microsoft documentation this variant of GetSystemTime guarantiees to get highest posible level of precision (<1us). https://learn.microsoft.com/en-us/windows/win32/api/sysinfoapi/nf-sysinfoapi-getsystemtimepreciseasfiletime


jzmaddock:
I'm closing this because:

1) This is probably best dealt with at a library level (ie not in Boost.Config), and
2) There are no Boost libraries actually using BOOST_HAS_GETSYSTEMTIMEASFILETIME anyway (other than the WinAPI wrappers which are using it correctly).

Please reopen if you can show a use case for a specific Boost library.

PR #531, opened 2025-12-11
Merge Boost.StaticAssert into Boost.Config.

Lastique:
This follows [this](https://lists.boost.org/archives/list/boost@lists.boost.org/thread/TWKBI5NXF772FGOBZA2FLIC2S7AM6OCQ/#TWKBI5NXF772FGOBZA2FLIC2S7AM6OCQ) discussion on the Boost ML.

This merges the sources, documentation, tests and examples of Boost.StaticAssert into Boost.Config. Documentation has been updated to be a part of Boost.Config documentation. One other change is use a more specific include from Boost.TypeTraits in static_assert_example_2. The rest is unchanged and at the state of Boost 1.90.0 release (https://github.com/boostorg/static_assert/tree/boost-1.90.0).

The test/cmake_install_test and test/cmake_subdir_test directories from Boost.StaticAssert are not included in this PR. These tests were not enabled since Travis CI went down and are not applicable to the merged Boost.StaticAssert as these tests check for whether Boost.StaticAssert is usable from CMake. Similar tests could be written for Boost.Config, but that would have to be a separate work.

This PR must be merged together with https://github.com/boostorg/static_assert/pull/18

### DIFF
--- a/doc/html/boost_config/acknowledgements.html
+++ b/doc/html/boost_config/acknowledgements.html
@@ -3,7 +3,7 @@
 <meta http-equiv="Content-Type" content="text/html; charset=UTF-8">
 <title>Acknowledgements</title>
 <link rel="stylesheet" href="../../../../../doc/src/boostbook.css" type="text/css">
-<meta name="generator" content="DocBook XSL Stylesheets V1.79.1">
+<meta name="generator" content="DocBook XSL Stylesheets Vsnapshot">
 <link rel="home" href="../index.html" title="Boost.Config">
 <link rel="up" href="../index.html" title="Boost.Config">
 <link rel="prev" href="rationale.html" title="Rationale">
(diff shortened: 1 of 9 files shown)


Lastique:
I'll see if I can extract with history, but I'm not sure it'll work if there are commits that touch paths that are present in Boost.Config.

pdimov:
I think we should delay this until the b2 msvc-14.5 update (hopefully soon) and the associated changes here in Config.

Lastique:
Thank you.

Lastique:
There's quite a bit of code in static_assert.hpp, and not everyone needs it. Eventually, the number of users will decrease as people switch to core `static_assert`. So I think it is better to stay separate, but I don't have a strong opinion on this.

pdimov:
Neither do I, to be honest. Most of the code is actually ifdef-ed out, but there's no pressing need to make any changes.

PR #532, opened 2026-01-04
Add BOOST_LIBSTDCXX_VERSION_WORKAROUND_GUARD

pdimov:
This avoids the `-Wundef` warning when `BOOST_WORKAROUND(BOOST_LIBSTDCXX_VERSION, ...)` is used without a `defined(BOOST_LIBSTDCXX_VERSION)` check.

### DIFF
--- a/include/boost/config/workaround.hpp
+++ b/include/boost/config/workaround.hpp
@@ -264,6 +264,11 @@
 #else
 #define BOOST_CLANG_VERSION_WORKAROUND_GUARD 0
 #endif
+#ifndef BOOST_LIBSTDCXX_VERSION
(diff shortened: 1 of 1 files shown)


pdimov:
Merging because the `-Wundef` warning is unavoidable on the user side even if guarded.

```
./boost/unordered/detail/type_traits.hpp:66:58: error: "BOOST_LIBSTDCXX_VERSION_WORKAROUND_GUARD" is not defined [-Werror=undef]
 #if defined(BOOST_LIBSTDCXX_VERSION) && BOOST_WORKAROUND(BOOST_LIBSTDCXX_VERSION, < 50000)
                                                          ^
```

PR #535, opened 2026-03-07
Adds support for C++20 modules

anarthal:
Adds "compatibility headers" <boost/config/std/xyz.hpp> to easily ifdef-out std dependencies in other libraries
Adds boost/config/disable_module_warnings.hpp, to disable warnings about angle-bracket includes in purviews
Changes some includes under some compilers to be more lightweight


anarthal:
See https://github.com/boostorg/cmake/pull/105 for the full discussion, https://github.com/anarthal/boost-cmake/blob/feature/cxx20-modules/modules.md for the rationale I've been following.

jzmaddock:
I'm not necessarily against this, but it will need documentation and testing support.

PR #540, opened 2026-07-08
Update names of environment variables as recommended by the latest CI…

jzmaddock:
… scripts.


jzmaddock:
@pdimov do you happen to know if depinst is compatible with being called on Boost.Config?  All the drone actions are failing with:

```
++ python tools/boostdep/depinst/depinst.py --include benchmark --include example --include examples --include tools config
Installing: config, headers, ../tools/boost_install, ../tools/build, ../tools/cmake
fatal: destination path '/drone/boost-root/libs/config' already exists and is not an empty directory.
fatal: clone of 'https://github.com/boostorg/config.git' into submodule path '/drone/boost-root/libs/config' failed
Failed to clone 'libs/config'. Retry scheduled
fatal: destination path '/drone/boost-root/libs/config' already exists and is not an empty directory.
fatal: clone of 'https://github.com/boostorg/config.git' into submodule path '/drone/boost-root/libs/config' failed
Failed to clone 'libs/config' a second time, aborting
```

Interestingly, our CI files do not call depinst anywhere, so I assume there must be a change in the drone scripts somewhere that's causing this @sdarwin ?

sdarwin:
@jzmaddock please merge PR 536

jzmaddock:
Ah, many thanks @sdarwin I knew I'd seen something relevant but not where...

pdimov:
Config is special-cased in depinst as an "essential" dependency and is always installed (it's used by the Jamfiles), which doesn't work that well when the library being depinst-alled is Config itself. :-)

I'll look into fixing this.

jzmaddock:
>I'll look into fixing this.

Thanks to Sam Darwin I have things working again, so don't worry too much!

pdimov:
I fixed it anyway.

PR #541, opened 2026-07-09
CI update 2026 2

jzmaddock:


jzmaddock:
@sdarwin do you happen to know if the drone failures:

```
+ ci/drone/linux-cxx-install.sh
/bin/sh: 60: ci/drone/linux-cxx-install.sh: Permission denied
```

are just random failures or a likely symptom of something else?

sdarwin:
The problem occurs before the error you mention.

`curl: (16) Error in the HTTP2 framing layer`

curl failed to download a file.   
Sporadic github.com outage.  
Retry the job.

Issue #499, opened 2024-06-22
msvc::no_unique_address


mglisse:
According to https://devblogs.microsoft.com/cppblog/msvc-cpp20-and-the-std-cpp20-switch/#c20-no_unique_address , MSVC ignores `[[no_unique_address]]`, but some versions provide `[[msvc::no_unique_address]]` and actually act on it (this has also been added to compatible compilers like clang-18+).
Would it make sense for BOOST_ATTRIBUTE_NO_UNIQUE_ADDRESS to expand to this attribute under the right conditions?

jzmaddock:
Sorry to be 2 years late to this party!

I'm not sure about this: doesn't the ABI breakages make this problematic?

mglisse:
Changing the ABI is the whole point of this attribute, it should be up to the user to decide if they want it or not, and if they use the macro, it means they want the different ABI. The user may be another boost library, that itself has users who may not be aware of this, so indeed it can cause problems I guess, but the existing BOOST_ATTRIBUTE_NO_UNIQUE_ADDRESS already causes binary incompatibility between the last version of gcc before the attribute was implemented and the next one.
Still, your choice, there is some sense in following Microsoft's choice of ~~penalizing their own users until they switch to a better compiler~~ favoring ABI compatibility at all cost (other compilers aren't much better there, ABI compatibility is killing C++).

pdimov:
FWIW, under msvc-14.2, this attribute works under C++14 and C++17, but causes a warning
```
C:\drone\boost-root\boost/unordered/detail/opt_storage.hpp(18): warning C4848: support for attribute 'msvc::no_unique_address' in C++17 and earlier is a vendor extension
```
which is a problem for `/WX` builds.

It's not clear whether the right course of action here is for every use to disable the warning (as we'll probably end up doing in Unordered), or not enable the attribute for msvc-14.2 under the earlier language modes.

msvc-14.3 doesn't seem to warn.

joaquintides:
IMHO the warning should be disabled here in Boost.Config, even if that can't be guarded with `warning(push)`/`warning(pop)` (which it can't)

pdimov:
This is a general warning, it's not specific to `msvc::no_unique_address`, and shouldn't be disabled globally.

pdimov:
There seem to be a lot of uses of BOOST_ATTRIBUTE_NO_UNIQUE_ADDRESS, so the most reasonable fix would probably be to not use `msvc::no_unique_address` on msvc-14.2.

[...]

PR #557, opened 2026-09-12
Disable the use of [[msvc::no_unique_address]] on versions prior to 1943…

pdimov:
…, because it causes a warning in C++14 and C++17 mode. Refs #499.


pdimov:
On second thought, updating msvc-14.3 silently changing the layout of classes using BOOST_ATTRIBUTE_NO_UNIQUE_ADDRESS might be problematic. We do this for different compiler versions (which is why we mangle the toolset name into the library name), but doing it within the same compiler version might be a bit too much.

